PR #141, opened 2014-12-28
Update modernizr

robbi5:
Hi,

your used version of Modernizr (1.6) triggers an GPU switch on every page load on macbooks with two gpus - see https://github.com/Modernizr/Modernizr/issues/1341

I looked around and found no usage of special Modernizr tests, so I created a custom build that only includes the core, html5shiv and the css classes.

Big thanks for the project :)
~robbi5


stefanw:
The .min extension was referenced but not used, that's why the build broke. Fixed that in aea4b483940f340dd3959de7ad2238ef4c6f4212 and merged to master.

PR #146, opened 2015-03-30
Use Whoosh instead of elasticsearch for tests; reducing dependencies for local testing

fin:
please?


stefanw:
Sure, good idea. Thanks for PEP8 fixes as well.

PR #147, opened 2015-04-16
Fix law assignment (for jurisdiction & public body)

fin:
fix Law/Jurisdiction / PublicBody assignment assignment issue.

By-Default, select law by priority; Meta adds 1000 to priority implicitly.
By-Default, select law by Jurisdiction. FoiLaw.laws becomes an override opportunity

Currently considering creating a script that compares the previous logic with the new one, and allows people to remove unnecessary overrides in publicbodies. Thoughts?

### DIFF
--- a/froide/foirequest/tests/factories.py
+++ b/froide/foirequest/tests/factories.py
@@ -278,7 +278,7 @@ def make_world():
         meta=True)
 
     for i in range(5):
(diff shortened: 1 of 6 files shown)


fin:
fixes #147

stefanw:
I'll have a look on the weekend, doesn't look to complicated though.

getjack:
@fin If the PR is still relevant could you fix the merge conflicts?

PR #155, opened 2015-08-29
fix initial user flow

fin:
- publicbody/fixtures/publicbody.json (we discarded publicbodytopic a while ago)
- syncdb does not have a --migrate flag anymore (with requirements.txt django)

$ ./manage.py syncdb --noinput --migrate
usage: manage.py syncdb [-h] [--version] [-v {0,1,2,3}] [--settings SETTINGS]
                        [--pythonpath PYTHONPATH] [--traceback] [--no-color]
                        [--noinput] [--no-initial-data] [--database DATABASE]
                        [--configuration CONFIGURATION]
manage.py syncdb: error: unrecognized arguments: --migrate


joelpurra:
@fin: thanks for the fix!

stefanw:
Thanks. Since Django 1.8 has deprecated `syncdb`, we should change this into `migrate` at some point.

PR #167, opened 2015-09-26
Allow users to change their names

LilithWittmann:
Sometimes users change their legal name and then it would be great if they could change their name easily on all their online accounts as well.
I simply added two fields to the UserChangeForm and now I (and all other people) can change their name :-)

### DIFF
--- a/froide/account/forms.py
+++ b/froide/account/forms.py
@@ -153,6 +153,16 @@ class PasswordResetForm(auth.forms.PasswordResetForm):
 
 
 class UserChangeForm(forms.Form):
+    first_name = forms.CharField(max_length=30,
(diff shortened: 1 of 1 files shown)


stefanw:
Cool, thanks! I will have a look into the tests.
I did not include the functionality originally because a change of name during the course of a request might confuse the public body and might open the door for abuse. I will discuss this change with the current community manager and may make it optional per deployment.

stefanw:
Hi, we decided against this I'm afraid. We had some experience with trolls and the ability to change the name by yourself would allow some behaviour in communication  with authorities we want to avoid. If you need a name changed, a email to support will do – or alternatively create a new account.

PR #179, opened 2016-03-19
User specific feeds

cbluoss:
This patch adds user specific feeds for requests made and requests followed.
Therefore added a uuid4 token to the user model for privacy.

README.md updated to Django 1.8 as required by requirements.txt

Suggestion:
update feed urls in foirequest/urls.py and use the regex to differ between rss and atom feeds (as I did).


yourcoke:
Be aware: The feed urls contain the user.id/pk.
I can't see any issues with this, but it can be removed in favor of a unique token.

Issue #194, opened 2016-05-31
Gettext issues on heroku


Alexander-Minyushkin:
In bin/post_compile you load file wich is not available, 

`GETTEXT_TARBALL=https://s3-eu-west-1.amazonaws.com/midgard-heroku/gettext.tar.gz`

```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AllAccessDisabled</Code><Message>All access to this object has been disabled</Message><RequestId>689A580334F43924</RequestId><HostId>WkT/QDZYseFdsx/EbYpafs3OIW11LDVQV5SjhSAoUO+Yz4DFxM6ikebKTT8T8XkHu1Gy7dtetBI=</HostId></Error>
```

Changing to latest version of gettext doesn't work either:

`GETTEXT_TARBALL=http://ftp.gnu.org/pub/gnu/gettext/gettext-latest.tar.gz`

```
remote: -----> Running post-compile hook
remote: -----> In post-compile hook
remote: -----> Installing gettext msgfmt...
remote: -----> done installing gettext msgfmt
remote: -----> Compiling translation files
remote:        CommandError: Can't find msgfmt. Make sure you have GNU gettext tools 0.15 or newer installed.
```

stefanw:
Interesting, I haven't deployed on Heroku in a while. There are a couple of options:
1) You don't care about translation, so ignore it and run in English
2) Commit compiled translation files (.mo) and push them to Heroku
3) Use a build pack of some sort, [try one of these maybe](https://elements.heroku.com/search/buildpacks?utf8=%E2%9C%93&q=gettext)

Sorry I can't be of more help as Heroku was just a demo deployment. If you figure out a good solution, please let me know!

stefanw:
You might also try to swap out the gettext URL with the one here:
https://github.com/lexifdev/heroku-buildpack-gettext/blob/master/bin/compile

The tarball should contain the gettext binaries compiled to run on the Heroku platform, so not the source tarball you linked to.

Alexander-Minyushkin:
Thanks! looks like GETTEXT_TARBALL=http://fb_wanted.s3.amazonaws.com/bp/gettext.tar.gz did the job
But now it compile all possible languages, not that defined in settings.LANGUAGES

stefanw:
Better than nothing ;) You can add the `-l` flag to compilemessages.

pdehaye:
The proposed fix works.

PR #195, opened 2016-05-31
Fix for "CommandError: Can't find msgfmt." on Heroku

Alexander-Minyushkin:
This fix discussed here https://github.com/stefanw/froide/issues/194

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -13,8 +13,8 @@ echo "-----> In post-compile hook"
 if [ ! -d $TARGET_DIR/gettext ]; then
     echo "-----> Installing gettext msgfmt..."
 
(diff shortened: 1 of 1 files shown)


pdehaye:
I encountered the same problem as #194 , and this fix worked for me too. Please accept the pull request.

stefanw:
I'm slightly nervous about this because it downloads a random binary and executes it with access to all Heroku credentials. The checksum just ensures that it is the same random binary every time ;).

Of course this is not your fault as it was like this before, but it stays an ugly hack that trusts a third party binary. If there was a public trusted source for gettext builds or existing buildpacks from trusted sources, that would be preferable.

stefanw:
We changed it so that gettext .mo files are commited as well, so no need for compiling on deploy anymore.

PR #203, opened 2016-08-17
Compare lowercase versions of domain names

pdehaye:
When accepting emails for delivery, we only compare lowercase versions of domain names. 
(do replies to inbound@fragdenstaat.de or inbound@FragDenStaat.de both get routed correctly?)


stefanw:
This is included in the refactor in 8b9dc25d9e6c790c984d885a66dd34f0193aae16

PR #204, opened 2016-08-17
Froide is not compatible with django 1.10

pdehaye:
Hence changed the requirements.txt file. This is due to the `patterns` module, deprecated in 1.8 already.

http://stackoverflow.com/questions/8074955/cannot-import-name-patterns

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pip==8.1.1
-Django<1.10
+Django==1.8
 Markdown==2.6.6
 celery==3.1.23
 django-haystack==2.5.dev1


pdehaye:
1.10 does not work because of http://stackoverflow.com/questions/8074955/cannot-import-name-patterns
1.8 does not work because of the `builtins` attribute for `overextends`: http://stackoverflow.com/questions/37509628/typeerror-with-django-1-8-13-but-not-with-django-1-9-6
In any case, my fix is wrong and it is not even clear there needs to be a fix. Closing this pull request.

pdehaye:
Actually, there needs to be a fix. Right now Froide-theme requires django 1.6.x, which means when you try out theming it messes up everything: installs older version of django, then Froide itself does not catch it.

stefanw:
I will try to have a look later this week. Please note that [fragdenstaat_de](https://github.com/okfde/fragdenstaat_de) (the theme project for FragDenStaat.de) runs on Django 1.9. Possibly have a look at the differences.

PR #205, opened 2016-09-22
Jurisdiction template refers to the wrong variable

pdehaye:


stefanw:
Indeed this is not the case. The sidebar in this template is supposed to convey general information about the whole site (e.g. [as can be seen for Berlin](https://fragdenstaat.de/berlin/)). However, I would agree that the jurisdiction templates need a do-over.

pdehaye:
My apologies, it was counter-intuitive from the information available in the Froide repo itself.

PR #215, opened 2017-07-15
Improve elasticsearch autocomplete

robbi5:
This PR adds a custom analyzer and tokenizer for autocompletion with elasticsearch. Boosts for things like Ministries can be configured in the settings


stefanw:
Has landed.

Issue #213, opened 2017-06-04
Sorting Messages correctly


arnese:
Now, when manually uploaded, sent and recieved messages via post get the time stamp "0:00". That can lead to them being posted in the wrong order if multiple messages from one day are uploaded. If possible, the time stamp should just be the current time so that messages get in the order that they are uploaded.

arnese:
solved with https://github.com/okfde/froide/pull/235

PR #235, opened 2018-03-06
Add current time to timestamp of manually uploaded messages

elya5:
This solves #213 by adding the current time to the date of the upload form.


jfilter:
Hey Elias, thanks for helping out! 👌

PR #241, opened 2018-04-08
Add missing jsonp dependency

rixx:


stefanw:
Thanks! I'm currently working with Pipenv. Managing to kinds of dependency files is not a good idea...

rixx:
Plus the ``setup.py``, yeah, that's annoying. Want to move the docs to use pipenv? (To coordinate the requirements.txt and the setup.py, you can put ``-e .`` as single line in the requirements.txt, but I don't think there is Pipfile integration yet anywhere.)

PR #262, opened 2019-01-13
[README]: correct to current naming / branding for Open Knowledge.

rufuspollock:


stefanw:
Waited just long enough so it's [Open Knowledge Foundation](https://okfn.org/) again. 😃

PR #285, opened 2019-04-02
Update download buttons in sidebar.html

Mineroboter:
Currently, the download buttons in the sidebar look like this:
buttons1

In order for it to look better, I added these changes, which stretch the buttons to 100 % width and add a small margin between the two buttons:
buttons21

But the translations need to be changed as well: "Download request as ZIP-file" is too long. I propose "Download as ZIP-file" and in German "Download als ZIP"; as well as "Download as PDF". Then the strings look more symmetric and still express the same information. But I don't know how to change these translations, so feel free to help.

### DIFF
--- a/froide/foirequest/templates/foirequest/snippets/sidebar.html
+++ b/froide/foirequest/templates/foirequest/snippets/sidebar.html
@@ -107,13 +107,13 @@
  {% if object|can_read_foirequest_authenticated:request %}
     <li class="nav-header">{% trans "Download" %}</li>
     <li class="nav-item">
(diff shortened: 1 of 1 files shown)


Mineroboter:
@stefanw Translations and both strings still need to be changed. Could you take care of that?

stefanw:
I added some more changes in 094fa74f5c682b6212562f6499352642eaca5ce9 and replaced `w-100` with `btn-block`. But thanks for kicking this off, the look of the sidebar annoyed me for some time. Will be deployed when the server load is back to normal.

PR #287, opened 2019-04-16
Use correct undo/redo icons in PDF redact

markuspoerschke:
Before:

grafik

After:

grafik

### DIFF
--- a/frontend/javascript/components/redaction/redaction.vue
+++ b/frontend/javascript/components/redaction/redaction.vue
@@ -32,11 +32,11 @@
       <div v-if="ready" class="btn-toolbar col-lg-12">
         <div class="btn-group mr-1">
           <button class="btn btn-light" @click="undo" :disabled="!canUndo" :title="i18n.undo">
-            <i class="fa fa-step-backward"></i>
+            <i class="fa fa-undo"></i>
           </button>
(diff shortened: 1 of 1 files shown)


robbi5:
Oh. I've talked with @arnese about these buttons a few days ago. The circle buttons would be a nice choice, but could also be interpreted as "rotate page". Not sure about it 🙃

jfilter:
Yeah, I also liked the old ones better. The current ones could also be interpret as 'go to next/prev. page'.

stefanw:
The Font Awesome icons (also official iOS icon) for undo/redo are very unfortunate as they look like rotate page icons. Because we have an actual rotate page use case on the new upload interface (likely in the logical step prior to redaction) this leads to real confusion. aa62398990f67f5d4cf89a5d2e277bd94d9bc58c changed the icons in an attempt at making them different but with a similar meaning: step forward / step back icons.

Text in English would be fine, but "Rückgängig" und "Wiederholen" on buttons also doesn't look very nice. The whole interface is not easily usable (close to impossible on touch screens), so the undo/redo icons are the smallest problem.

I would appreciate help with getting the whole redaction interface into a better state. But please let's not spend time bikeshedding the icons there.

PR #379, opened 2020-10-25
Fix class name

alehaa:
FOI mail class name changed in 3de57145e7f5c32aa3ab8db9785b2a7cfdcb16ce.


stefanw:
Thanks! Went unnoticed because we are not using the management command.

PR #398, opened 2020-12-15
change used genderstyle to use colons; change to "Damen bis Herren" in Greetings

### LINKED ISSUE #397: german translation: augment used version of gendering to include non-binary people
The current german translation uses a slash "/" to seperate a male associated and a female associated version of the same function of a person. This implies a choice from all *listed* options, thereby excluding non-binary people. To fix this I propose changing it to a colon ":", as it is much more naturally to follow with screen readers than an asterisk "*" since it is being read as just the short gap as intended in spoken language, but still is not associated with concluded lists of options. An example of the resulting language would be "Antifaschist:in"
Further I'd like to change the standard greeting subtly to "Sehr geehrte Damen bis Herren" (from currently Damen *und* Herren") to make clear that men and women are not the only genders to be observed while fitting nicely with language "flow" (I lack a better word here). This is however against a conception of gendering being supposed to point people to the fact that it is being actively augmented to include everyone and thereby pushing towards reflection of gender roles as percieved within society.

This issue is to start a discussion on the topic if needed and before starting the actual work see, if there are aspects I missed.

Unip0rn:
fixes #397 
I think I got them all, if there are relevant files outside of /locale I am not aware of them and they probably still to do.

### DIFF
--- a/locale/de/LC_MESSAGES/django.po
+++ b/locale/de/LC_MESSAGES/django.po
@@ -129,31 +129,31 @@ msgstr "Sessions gelöscht."
 
 #: froide/account/admin.py:167
 msgid "Delete sessions of users"
-msgstr "Sessions des Nutzers löschen"
+msgstr "Sessions des:der Nutzer:in löschen"
 
 #: froide/account/admin.py:172
 msgid "Users canceled."
-msgstr "Nutzer gekündigt."
+msgstr "Nutzer:in gekündigt."
 
 #: froide/account/admin.py:174
 msgid "Cancel account of users"
-msgstr "Nutzerkonto kündigen"
+msgstr "Nutzer:innenkonto kündigen"
 
 #: froide/account/admin.py:179
 msgid "Users logged out, deactivated and blocked."
-msgstr "Nutzer wurden ausgeloggt, deaktiviert und blockiert."
+msgstr "Nutzer:in wurden ausgeloggt, deaktiviert und blockiert."
 
 #: froide/account/admin.py:181
 msgid "Deactivate and block users"
-msgstr "Nutzer deaktivieren und blocken"
+msgstr "Nutzer:in deaktivieren und blocken"
 
 #: froide/account/admin.py:188
 msgid "User made private."
-msgstr "Nutzer wurde auf privat geschaltet."
+msgstr "Nutzer:in wurde auf privat geschaltet."
 
 #: froide/account/admin.py:190
 msgid "Make user private"
-msgstr "Nutzer auf privat schalten"
+msgstr "Nutzer:in auf privat schalten"
 
 #: froide/account/admin.py:195
 msgid "Can only merge two accounts at a time."
@@ -161,7 +161,7 @@ msgstr "Kann nur zwei Konten zusammenführen."
 
 #: froide/account/admin.py:205
 msgid "Account merging started..."
-msgstr "Nutzerkonto-Zusammenführung angestoßen..."
+msgstr "Nutzer:innenkonto-Zusammenführung angestoßen..."
 
 #: froide/account/admin.py:206
 msgid "Merge accounts (keep older)"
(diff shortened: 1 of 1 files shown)


locale/de/LC_MESSAGES/django.po
```diff
@@ -9610,7 +9610,7 @@ msgstr ""
 "\n"
 "An: Bundesministerium des Innern\n"
 "\n"
-"Sehr geehrte Damen und Herren,\n"
+"Sehr geehrte Damen bis Herren,\n"
```
rugk, line 9613: I guess agencies will be confused by that…

rugk:
BTW is it intentionally that you committed with different mail addresses, @Unip0rn?

alehaa:
I don't know if this is the right place for a discussion, but I would prefer to entirely remove genders to avoid pitfalls, if they're not required.

```diff 
-Bitte setzen Sie jetzt ein Passwort für Ihr Nutzerkonto.
+Bitte setzen Sie jetzt ein Passwort für Ihr Konto.
```

locale/de/LC_MESSAGES/django.po
Unip0rn: I don't think the confusion would last very long and the alternatives are more invasive to the "flow". E.g. "Sehr geehrte Menschen" may even be read as derogatory, a mere "Hallo" seems clearly too informal to me. Also when it causes confusion, that probably is for expanding the recipients horizont to people beyond the gender binary and therefor would be considered a feature, not a bug. I am however open to suggestions.

Unip0rn:
> BTW is it intentionally that you committed with different mail addresses, @Unip0rn?

@rugk I didn't. The fixing plural thing was a maintainer's work.

locale/de/LC_MESSAGES/django.po
KayleeTheMech: `Sehr geehrte Alle,\n?`

rugk:
@Unip0rn Please do also note one _important_ precautionary warning of me if you consider these email addresses private: In case you did not notice, you have committed (i.e. created commits) with potentially sensitive data like your (full/private) name or mail.
Anyone, who clones your fork/this repo can view these private details, exactly as you've committed them. Also, if this PR gets merged, this will stay in the (commit) history of this project and will thus hard to be removed later.

So if this was unintentional and you want that data to be changed/removed first, better do it now. :smiley: [Here is a good resource](https://stackoverflow.com/a/750182/5008962) ([actually from GitHub's help](https://help.github.com/en/articles/changing-author-info)) on how to change your name/mail afterwards.
Also note that when setting your git [username](https://help.github.com/en/articles/setting-your-username-in-git) and [especially your email](https://help.github.com/en/articles/setting-your-commit-email-address), you can actually write/choose anything there. As for GitHub, it is recommend to use your GitHub username and – if you want to keep your mail private – the GitHub provided `noreply` address. You can [even block pushes of commits with private mails](https://help.github.com/en/articles/blocking-command-line-pushes-that-expose-your-personal-email-address).

So far just the possibilities. I just wanted to serve this as a friendly reminder, so you don't notice this only when it's too late. What you'll do depends solely on you. :smile:

locale/de/LC_MESSAGES/django.po
Unip0rn: > `Sehr geehrte Alle,\n?`

Sounds good to me, good idea! :)

rugk:
Ah sorry forget my previous comment, I was confused by two usernames who committed.

Unip0rn:
no harm done, but thx for having a look-out and for the links, might come in handy :)

pajowu:
Hey, first of all let me thank you a lot for creating this PR. I just started working at FragDenStaat this week and wanted to do something similar before noticing this 😃 

We want to tackle this issue soon, however this PR contains two things:

1. Gender-neutral language towards the users
2. Gender-neutral greetings in emails to public bodies

Gender neutral language towards the users could be tackled right away and is on my todo-list for this or next week.

For the second point, we still need to have some discussion within the team on how we want to solve this.

Therefore I'd like to split this PR into two. This way we can merge parts of it (solving the first point) soon. 

Since you already opened the PR: you want to do this?
Otherwise I would be happy to do so in the next days

pajowu:
I opened a new PR #509 with the changes towards the users. I manually checked all strings again for that, but please feel free to have a look as well

locale/de/LC_MESSAGES/django.po
codedust: How about `Guten Tag`?

```suggestion
"Guten Tag,\n"
```

pajowu:
To keep everyone interested updated: Changing the greeting-style is something we are interested in, but if done properly would require some more work.

If we change it, a good way would be adding an option to choose the specific greeting you want to use. However the greetings are not only set in the translations, but also in texts in database-models (for example `FoiLaw.letter_start`). So a good solution would need to also take care of that.

PR #410, opened 2020-12-29
Fix long request text fade-out

alehaa:
This PR fixes the text fade-out on the request page (alpha) for Safari to not fade to gray, but the background color instead. Firefox should be unaffected by this patch and still fades correctly.

screenshot


alehaa:
@stefanw is there any chance to merge this PR, now the new request page design has become beta?

PR #421, opened 2021-02-22
Redesign request sent page

krmax44:
See reference Figma: https://www.figma.com/file/E1b5GfyjdOzlpSjAR6eov0/Anfrage-versendet?node-id=0%3A1 and okfde/fragdenstaat_de#205


stefanw:
I merged it locally without the additional po/mo files. I try to keep translations in separate commits and only build de language. That allows easier merging (.po files are a pain to merge). Additional languages are only generated on release, because it triggers translator notices on transifex (or it used to, haven't checked in a while).

PR #424, opened 2021-03-06
NVC translation for reminder email

lanmarc77:
NVC= non violent communication
There are some authorities that find the current text of the (german) reminder email not appropriate. I tried to translate it into non violent communication. This should still reflect the lawfully binding wish of the writer so should not change anything from the law side but might help some authorities in understanding the other side as well.
Based on this discussion: https://forum.okfn.de/t/antwort-der-behoerde-auf-erinnerungsnachricht-gleichwohl-moechte-ich-sie-bitten-ihr-anliegen-in-einem-angemessenen-ton-vorzutragen/


pajowu:
Thank you very much for the suggestion, however (as discussed in the forum) we decided against using this text for now.

PR #430, opened 2021-03-12
Refresh FoiLaw page

krmax44:
new features:
- `legal_text` field for FoiLaw, allowing to specify the law in its entirety
- anchor links: `/gesetz/ifg-berlin/#§3` will link to § 3 of the law
- refreshed law page, preview (with and without specified legal text):

Screenshot from 2021-03-12 16-41-36
Screenshot from 2021-03-12 16-41-29

I'm struggling with the Django i18n system. Running `manage makemessages` doesn't update the changed blocktrans strings somehow. :thinking: That's why translations are still missing.


krmax44:
Maybe this would be a good opportunity to add a Gebührenordnung field as well? Looks a bit lost like that: https://fragdenstaat.de/gesetz/ifg-nrw/

stefanw:
You need to go into the Python package dir (here froide_legalaction) with activated virtualenv and run django-admin makemessages -l de. That will then detect the locale dir present there and update the .po file. Feel free to document that!

Issue #464, opened 2021-11-06
Maximum number of messages within timeframe incorrect


teriky:
When sending the 6th message within a timeframe of 24h one got the message "Sie haben die Begrenzung von 6 Anfragen in 6 Stunden überschritten. Dies ist eine automatische Vorsichtsmaßnahme."

This is not correct because the message is displayed after sending 5 messages within 24h: 5 messages sent within 6 hours, the 6th is rejected).

A similar error occures when sending the 8th message within 24  hours (7 messages sent, the 8th is rejected).

stefanw:
Sorry, I don't quite understand the issue.

> When sending the 6th message within a timeframe of 24h one got the message
> ...
> This is not correct because the message is displayed after sending 5 messages within 24h

Is it the 6th or the 5th message? What did you do, what message did you get and what message do you think you should get?

teriky:
Steps to reproduce:

- precondition: no messages sent within the last 24 hours
- send 5 messages within 1 hour up to 6 hours
- send the 6th message within the same 1 hour up to 6 hours

_Expected behaviour:_ the 6th message is sent, no hint is displayed

_Actual behaviour:_ the 6th message is **not** sent, the hint "You have exceeded the limit of 6 requests in 6 hours. This is an automatic precaution." is displayed.

![hint message](https://user-images.githubusercontent.com/40558729/140653638-053eacb7-9b77-47e5-81ad-c57b20352171.jpg)

The same behaviour occurs with the limitation of 8 messages per 24 hours: Only 7 messages within 24 hours can be sent.

stefanw:
I think one confusion point is that there were no distinct throttle messages for sending requests and sending *messages* in requests. [This is fixed](1049b4f7402f2b18625c84f6a5576eb2a5525130).

[The throttle configs for requests and messages are distinct and are documented here](https://github.com/okfde/fragdenstaat_de/blob/master/fragdenstaat_de/settings/base.py#L519-L529).

The fail condition is [`message_count + 1 > fail_count`](https://github.com/okfde/froide/blob/main/froide/foirequest/utils.py#L46) which seems correct to me. I also see you closed your PR again.

[There's a basic test for request throttling](https://github.com/okfde/froide/blob/6ebed3fced319755956113b1cd58f1d22ce6c949/froide/foirequest/tests/test_request.py#L1781). Maybe you can come up with a failing test for messages?

PR #465, opened 2021-11-11
Fix off by one error in throttle

teriky:
The message throttling starts too early if 1 is added to the number of messages in the requested period.

@stefanw: This is a proposal for #464.


teriky:
My fix defined an off by two error.

PR #468, opened 2021-12-15
Allow TESSERACT_LANGUAGE to be overridden in settings

fin:
At FragDenStaat.at, we're using the language code de-at to allow us to use the German translations while overwriting some messages. This throws off Tesseract / django-filer. I'd like to propose allowing a seperate setting for the tesseract language code.


stefanw:
Thanks!

PR #480, opened 2022-02-07
Add Two-factor auth

### LINKED ISSUE #434: feature request: two-factor authentication
Whoever has access to a froide user account can send letters to any government agency under the name of the user account's holder. This is especially problematic for FOI requests that are automatically signed with a signature previously uploaded by the user, which is done when froide automatically sends a fax instead of an email for authorities that require requests to be submitted in "written form" (which is a cool feature, by the way!).

I therefore believe that it would make sense to introduce (opt-in) two-factor authentication, to decrease the risk that anyone could gain access to another user's account and send out signed letters to government agencies, which, in turn, would probably not question the authenticity of such a message immediately, given the fact that it was issued via FragDenStaat.

I'm not arguing for 2FA via SMS here - that would be relatively expensive, besides not really being safe in the first place. Something like [TOTP](https://en.wikipedia.org/wiki/Time-based_One-Time_Password) on the other hand, which is used by many popular services, might be relatively easy to integrate, since suitable django plugins probably already exist (?).

*PS: Thank you for building and maintaining this service :green_heart:*

stefanw:
Using https://github.com/xi/django-mfa3 as basis.

Closes #434


froide/templates/mfa/auth_recovery.html
```diff
@@ -0,0 +1,53 @@
...
+      </a>
+    {% endif %}
+    {% if mfa_keys.by_method.recovery %}
+      <a class="btn btn-light" href="{% url 'mfa:auth' 'recovery' %}">
+        {% translate "Use recovery code" %}
+      </a>
+    {% endif %}
+  </p>
+</form>
+{% endblock %}
```
rugk, line 53:
```suggestion
{% endblock %}
```

all other files have a line-ending (is your `editorconfig` /editor properly setup - if you have one?)

froide/templates/mfa/create_TOTP.html
```diff
@@ -0,0 +1,35 @@
...
+{% block title %}{% translate "Setup authenticator app" %}{% endblock %}
+
+{% block mfa_heading %}{% translate "Setup authenticator app" %}{% endblock %}
+
+
+{% block mfa_inner %}
+
+<ol class="mt-3">
+  <li>
+    {% translate "Scan the code with your authenticator app (e.g. Google Authenticator or Authy)." %}
```
rugk, line 16:
```suggestion
    {% translate "Scan the code with your authenticator app (e.g. Google Authenticator, Aegis or Authy)." %}
```

Let's maybe mention at least one FLOSS alternative… :upside_down_face:

rugk (commented):
I have no experience with Python and so on, but [the Readme of the project states](https://github.com/xi/django-mfa3) you should “use django-mfa3 with [django-axes](https://github.com/jazzband/django-axes) for rate limiting. It is also compatible with [django-stronghold](https://github.com/mgrouchy/django-stronghold/).” I don't see this done/being used anywhere.

So is there any rate limiting implemented to prevent brute-force attempts here? If not, it is likely a good idea to do so.

cf. https://github.com/xi/django-mfa3/issues/3

stefanw:
We are currently doing throttling with a web server configuration. Upgrading to something more app specific may be worth it.

Issue #378, opened 2020-10-23
Bug when uploading same file (name) twice


rugk:
Some bugs related to uploading the same file twice… :upside_down_face: 

![funny jeopardy "daily double" animation. Daily Double GIF by Jeopardy!](https://media.giphy.com/media/2yvoIFyZghBDszbIk3/giphy.gif)

(Sorry, I could not resist. :see_no_evil:)

## STR: Double file upload

0. Attach a new "received postal mail".
1. Upload a file named `test.pdf`.
2. Upload another (the same) file named `test.pdf`.

**What happens:**
> Wird geladen... Dieses Dokument wird gerade erstellt. Dies kann mehrere Minuten dauern. test.pdf 

This is shown for an infinite time.

**What should happen:** Maybe show me an error that the same file (name) already exists? Or automatically rename?

## STR: Double file upload and redact

0. Attach a new "received postal mail".
1. Upload a file named `test.pdf`.
2. Redact that file, so you get a file named `test_geschwaerzt.pdf`.
3. Upload another (the same) file named `test.pdf`.

**What happens:**
It is successfully uploaded.

**What should happen:** Maybe show me an error that the same file (name) already exists? Or automatically rename?

## STR: Double file upload, redact and delete

0. Attach a new "received postal mail".
1. Upload a file named `test.pdf`.
2. Redact that file, so you get a file named `test_geschwaerzt.pdf`.
3. Upload another (the same) file named `test.pdf`.
4. Now (noticing the error), you delete (the latest upload of) `test.pdf`.

**What happens:**
It deletes both `test.pdf` and `test.pdf` (i.e. both files).
They are likely referenced by their file name.

**What should happen:** Reference files by some ID, not their file name?

moritzgloeckl:
Hi @arnese @rugk,

I've been looking into these issues to take a look at the code base and to be able to contribute to this great project. While investigating, I found out that there's actually a built-in functionality to be able to replace uploaded files if their filename matches. Once I got around the endless progress (pending attribute was not set correctly) I've tested this functionality, however, it doesn't work as expected (not replacing the existing file). Before going deeper into this, I'd like to clarify the expected behaviour if a user uploads an already existing file:

1. Replacing the current file

OR

2. As @rugk suggested, renaming the files, e.g. `_1`, `_2`, `_3`

[...]

PR #504, opened 2022-04-13
 #378 Don't replace same named files but upload the new file with the filename numbered in attachments

cafischer:


froide/foirequest/tests/test_helper.py
```diff
@@ -0,0 +1,39 @@
+from django.test import TestCase
+
+from froide.foirequest.models import FoiAttachment
+from froide.helper.storage import get_numbered_filename
+
+
+class HelperTest(TestCase):
```
cafischer, line 7: This should have been a unit test not needing the django stuff, but I did not find any so far. So I wasn't sure whether and how to set it up... What do you propose?

stefanw:
Currently reviewing this. Our codebase is a bit messy in regards to making filenames on attachments unique.

PR #455, opened 2021-08-11
Feature rework mail delivery reporting

xenein:
this a first draft for #454

PR #506, opened 2022-04-20
Add postfix logparsing as a seperate task

### LINKED ISSUE #454: Rework mail delivery reporting from pull to push
Currently, every sent FOI mail puts a task on the queue that checks mail.log for delivery status of that mail. If not found, the task is retried ([code here](https://github.com/okfde/froide/blob/main/froide/foirequest/models/message.py#L512)) with exponential backoff.

This leads to many tasks doing the same work parsing mail.log. Instead, there should be either be:
- a process that follows the log for delivery notifications (e.g. a continuously running Django management command)
- a celery beat task that checks the log for new entries every minute or so (easier to deploy)

Both need to follow the log ([maybe with this](https://github.com/bgreenlee/pygtail)) and need to store state somewhere to know what they have last seen and handle log rotation. Maybe the `bounce` app is the right place for that code.

Delivery events (success, failure, pending) with log lines need to be sent (maybe via Django signal) and processed by FoiMessage app to create its DeliveryStatus objects.

Currently, the details of how delivery checking works is configurable, maybe it can stay that way, even though right now we are only parsing Postfix logs: https://github.com/okfde/froide/blob/main/froide/foirequest/delivery.py

pajowu:
This continues @xenein‌s work in #455 and closes #454 

This adds a new [regex-based postfix log parser](https://github.com/okfde/froide/blob/66ac22f77d9142e1962a99cc0878aa3461c58f59/froide/helper/email_log_parsing.py#L13) which sends a [`froide.helper.signals.email_left_queue`](https://github.com/okfde/froide/blob/66ac22f77d9142e1962a99cc0878aa3461c58f59/froide/helper/email_log_parsing.py#L174) signal for each message which was removed from the postfix queue (mostly because they are delivered), which is then used to [update the message status](https://github.com/okfde/froide/blob/66ac22f77d9142e1962a99cc0878aa3461c58f59/froide/foirequest/signals.py#L464) 

Since the postfix parser is no longer specific to foi messages, I moved the code to `froide.helper`.

### DIFF
--- a/froide/helper/tasks.py
+++ b/froide/helper/tasks.py
@@ -8,6 +8,7 @@ from elasticsearch.exceptions import ConnectionTimeout
 
 from froide.celery import app as celery_app
 from froide.foirequest.models.request import FoiRequest
+from froide.helper.email_log_parsing import check_delivery_from_log
 
 logger = logging.getLogger(__name__)
 
@@ -52,3 +53,8 @@ def search_instance_delete(model_name: str, pk: Optional[int]) -> None:
     instance.pk = pk
     instance.id = pk
     registry.delete(instance, raise_on_error=False)
+
+
+@celery_app.task(expires=60 * 60)
+def check_mail_log():
+    check_delivery_from_log()
--- a/froide/helper/tasks.py
+++ b/froide/helper/tasks.py
@@ -55,6 +55,6 @@ def search_instance_delete(model_name: str, pk: Optional[int]) -> None:
     registry.delete(instance, raise_on_error=False)
 
 
-@celery_app.task(expires=60 * 60)
+@celery_app.task(expires=60)
 def check_mail_log():
     check_delivery_from_log()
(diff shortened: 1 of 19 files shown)


froide/helper/tasks.py
```diff
@@ -52,3 +53,8 @@ def search_instance_delete(model_name: str, pk: Optional[int]) -> None:
     instance.pk = pk
     instance.id = pk
     registry.delete(instance, raise_on_error=False)
+
+
+@celery_app.task(expires=60 * 60)
```
stefanw, line 58: Expiry of 1 hour is likely too much. The task would run at least every couple of minutes or so. The idea is that the task cannot be in the queue for longer than it's scheduled frequency so it doesn't fill up the queue or gets executed in parallel.

pajowu: 👍

PR #510, opened 2022-04-20
🐛 Use bleach.linkify instead of urlize

### LINKED ISSUE #508: Quotes rendered as HTML code
Under some circumstances quotes are rendered with the corresponding HTML code instead of their visual.

**Steps to reproduce**

1. Edit the message of an uploaded document.
2. Copy the text following text to the letter summary:
```
'@ with spaces'
"@ with spaces"
'@NoSpaces'
"@NoSpaces"
'@No.spacesAndDot'
"@No.spacesAndDot"
```
3. Save the change.

_Expected result:_

The letter summary is displayed as entered.

_Actual result:_

snapshot

pajowu:
Froide converts urls in messages to clickable links, but not email adresses. Until now this was done by using djangos urlize[1] function and removing mailto links afterwards. However urlize should not be used on text that contains html or email adresses that contain single quotes (see note in [1]). This leads to the problem reported in #508.

This switches to bleach.linkify, which can handle existing html code and has an option to ignore emails.

[1]: https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#urlize

Closes #508


MagdaN (approved):
Works nicely (have not tested the new requirements thought) and looks good, thanks!

pajowu:
At least for FDS there is no new requirement, because we already depend on bleach due to `django-legal-advice-builder`

PR #512, opened 2022-04-23
✨ Add setting to change project of a request

pajowu:
**Note**: This contains a lot of similar code to the 'Add team to project' feature.

### DIFF
--- a/froide/foirequest/forms/__init__.py
+++ b/froide/foirequest/forms/__init__.py
@@ -14,7 +14,7 @@ from .message import (
     get_send_message_form,
 )
 from .postal import PostalUploadForm
-from .project import MakeProjectPublicForm
+from .project import AssignProjectForm, MakeProjectPublicForm
 from .request import (
     ApplyModerationForm,
     ConcreteLawForm,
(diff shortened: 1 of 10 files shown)


stefanw:
LGTM! I wouldn't worry about the duplication until the third time. And even then if it's more explicit and easier to follow some duplication can have benefits.

We don't have a good flow for working with translations, but generally I would only generate translations for our immediate use case. There is `make messagesde` which does some ignores and limits translation files to `de`. This keeps the PR smaller. Generating all translation files can be done for a proper release.

stefanw (approved):

stefanw (changes requested):
We should probably hide the form when the person has no projects to choose from (does not look like that's happening?). Currently majority of user cannot create projects and for them this form is not relevant.

MagdaN (commented):
Looks good and works nicely, the only problem I found is that when one of the options in the select field gets too long, the select field breaks the design.

pajowu:
I added a check if at least one project can be selected in the form, otherwise the setting is not shown

stefanw (approved):

PR #513, opened 2022-04-25
✨ Allow inline-editing of document details

pajowu:


froide/document/templates/filingcabinet/_base_set_form.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+
+<div class="mb-3 d-none" id="{{ form_id }}" data-autofocus="{%block autofocus%}{%endblock%}">
+  <form method="post" class="{% block fromclasses %}disable-submit{% endblock %}" action="{%block saveurl %}{%endblock%}">
```
stefanw, line 4: We don't have a style guide for template tags, but maybe we need one! We usually use spacing like in the Django docs: `{% block name %}`

froide/document/templates/filingcabinet/_set_description_form.html
```diff
@@ -0,0 +1,8 @@
+{% extends './_base_set_form.html' %}
```
stefanw, line 1: We have not used relative template paths anywhere yet as far as I'm aware, but I can see their use case.

froide/document/views.py
```diff
@@ -148,3 +151,37 @@ def upload_documents(request):
...
+def set_description(request, document):
+    return set_property(document, request, "description")
+
+
+def set_property(document, request, name):
+    value = request.POST.get(name, None)
+    if value is None:
+        return render_400(request)
+    setattr(document, name, value)
+    document.save()
```
stefanw, line 184: Here you could use [`update_fields` keyword argument](https://docs.djangoproject.com/en/4.0/ref/models/instances/#ref-models-update-fields) because you are really only dealing with one field.

stefanw (approved):
Looks good, small nitpicks. I like that we have now a generic way for inline editing, exactly what I wanted.

froide/document/templates/filingcabinet/_base_set_form.html
pajowu: Ooops, missed that

froide/document/views.py
pajowu: Nice, I didn't know about this option

PR #523, opened 2022-05-09
🩹 Add management command to pre-calculate redaction markup

pajowu:
Sometimes, especially for longer texts, calculating the redaction markup takes too long. We cache the results, but if the request is killed before the end of the calculation, it is never written to the cache.

This commit adds a management command that pre-calculates the redaction markup for all longer messages.


stefanw:
LGTM!

PR #533, opened 2022-05-27
Move link from title to attachment tile

### LINKED ISSUE #532: Add link to attachment tile not on attachment name
Anhang_Textlink

MagdaN:
fixes #532 

Not totally sure if this does not break something, e.g. if there are edge cases where there is a link on a linked tile


stefanw:
Unfortunately, that has resulted in a visual regression for some attachments, likely due to `div`s in `a`s and the presence of an `a` inside an `a`. I reverted in main. Not urgent.

Screenshot 2022-06-10 at 09 15 11

PR #538, opened 2022-06-17
🚸 Improve moderation of attachments

pajowu:
* Add option to mark as moderated
* Show link to message


stefanw:
Sieht nach fehlender Migration Datei aus.

pajowu:
So, jetzt ist sie da und auch richtig formattiert

frontend/javascript/components/moderation/moderation-attachments.vue
```diff
@@ -61,11 +81,25 @@ export default {
...
   },
-  methods: {}
+  methods: {
+    markModerated(att) {
+      postData(
+        getRedactionUrl(this.config.url.mark_attachment_as_moderated, att),
+        {},
+        this.$root.csrfToken
+      ).then(() => {
+        const index = this.attachments.findIndex((x) => x.id === att.id)
```
stefanw, line 96: Could this be a `this.attachments = this.attachments.filter(...)`?

froide/foirequest/views/message.py
```diff
@@ -608,3 +609,13 @@ def resend_message(request, foirequest, message_id):
 
     messages.add_message(request, messages.SUCCESS, _("The message has been re-sent."))
     return redirect(sent_message)
+
+
+def message_shortlink(request, obj_id, url_path=""):
```
stefanw, line 614: `url_path` seems unused. Maybe remove for simplicity.

froide/foirequest/urls/request_urls.py
```diff
@@ -203,4 +210,6 @@
     # Feed
     path("<slug:slug>/feed/", FoiRequestFeedAtom(), name="foirequest-feed_atom"),
     path("<slug:slug>/rss/", FoiRequestFeed(), name="foirequest-feed"),
+    # Shortlink for message
+    path("m/<int:obj_id>/", message_shortlink, name="foirequest-message_shortlink"),
```
stefanw, line 214: I see what you did there. It sure comes in handy. Our other more or less 'official' shortlinks (`/r/` and `/b/` etc.) are translated (in German to `/a/` and `/b/`), but we can keep it at a more low profile as well.

I would move this up above the `<slug:slug>` prefix to the other short link URL for some order and to match before slug (unlikely) matches the 'm' prefix).

froide/foirequest/models/attachment.py
```diff
@@ -163,6 +163,7 @@ class FoiAttachment(models.Model):
     is_converted = models.BooleanField(_("Is converted"), default=False)
     timestamp = models.DateTimeField(null=True, default=timezone.now)
     pending = models.BooleanField(default=False)
+    has_been_moderated = models.BooleanField(_("Has been moderated"), default=False)
```
stefanw, line 166: I would rather see this as `is_moderated` to keep in line with `is_converted` and `is_redacted`.

stefanw (changes requested):

frontend/javascript/components/moderation/moderation-attachments.vue
pajowu: No, because we cannot directly mutate properties. It seems to be considered a bad practice, can lead to vue missing updated and is forbidden by eslint rule 'vue/no-mutating-props'

stefanw (approved):

PR #542, opened 2022-06-20
✨ Allow finetuning of actions needed for ModerationTrigger to be applied

pajowu:
At the moment, triggers that send an email are *always* considered not applied, because SendUserEmail.is_applied always returns False. This PR allows more control over which actions are needed for a trigger to be considered applied, thus allowing us to ignore the SendUserEmail action

### DIFF
--- a/froide/foirequest/moderation.py
+++ b/froide/foirequest/moderation.py
@@ -175,7 +175,11 @@ def get_moderation_triggers(
             resolve_moderation_action(*action_args)
             for action_args in trigger["actions"]
         ]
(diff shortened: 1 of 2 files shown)


stefanw:
Vielleicht reicht auch das `all` durch ein `any` zu ersetzen?

pajowu:
["Depublishing"](https://github.com/okfde/fragdenstaat_de/blob/4dbf0f59d34da342473fbf2ed8e72c465785268c/fragdenstaat_de/settings/base.py#L586) would be an example: I think we want to check that both the request is published and the tag is added to the user, but do not care about the rest.

PR #543, opened 2022-06-20
🚸 Pre-fill name field in comment form

pajowu:


froide/comments/templatetags/froide_comment_tags.py
```diff
@@ -0,0 +1,11 @@
...
+
+from ..forms import CommentForm
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def get_froide_comment_form(context, object):
+    request = context["request"]
+    return CommentForm(object, initial={"name": request.user.get_full_name()})
```
stefanw, line 11: This assumes `request.user.is_authenticated` (Django anon user does not have `get_full_name()`). Right now we show comment forms only to logged in users, but this should be less assuming.

stefanw (changes requested):

PR #546, opened 2022-06-20
✨ Show warning for non meaningful subjects

### LINKED ISSUE #545: Force meaningful subject
Sometimes, subjects for FOI requests are pretty useless - for example, some people simply write "Anfrage" into the subject line. We should curate a list of subjects that are not permissible (while still allowing the word "Anfrage" in this case, of course).

pajowu:
Closes #545

### DIFF
--- a/froide/foirequest/views/make_request.py
+++ b/froide/foirequest/views/make_request.py
@@ -133,6 +133,9 @@ class MakeRequestView(FormView):
             "settings": {
                 "user_can_hide_web": settings.FROIDE_CONFIG.get("user_can_hide_web"),
                 "user_can_create_batch": self.can_create_batch(),
+                "non_meaningful_subject_regex": settings.FROIDE_CONFIG.get(
+                    "non_meaningful_subject_regex", []
+                ),
             },
             "url": {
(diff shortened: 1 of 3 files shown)


froide/foirequest/views/make_request.py
```diff
@@ -133,6 +133,9 @@ def get_js_context(self):
             "settings": {
                 "user_can_hide_web": settings.FROIDE_CONFIG.get("user_can_hide_web"),
                 "user_can_create_batch": self.can_create_batch(),
+                "non_meaningful_subject_regex": settings.FROIDE_CONFIG.get(
```
stefanw, line 136: This should have a default value of `[]` so the JS `.map` doesn't crash on `null` value.

stefanw (approved):

stefanw:
Generally, I would leave translation files out of PRs. They are a pain to conflict merge (at least I have not found a good way). I try to add translations in a separate commit on main before deploy.

froide/foirequest/views/make_request.py
pajowu: ✅

pajowu:
I added the default and removed the translation file

PR #547, opened 2022-06-22
🐛 Fix bug in toggleSlide not showing elements if their parent is hidden on page load

pajowu:
This also adds a selenium test for setting the request status


stefanw (approved):
Aaaaah, how is slide toggling so hard? Are we better off using a third party solution? Or should we replace that animation with something simpler? LGTM for now though.

Issue #515, opened 2022-04-26
Check validity of parsed email-adresses


pajowu:
Sometimes public bodies send email headers which do not conform strictly to rfc2822, which can lead to invalid email-addresses being returned by our email parser. We could filter the parsed adresses with some basic validity checks (e.g. Contains an `@`).

One example (adopted from a real email):

``` python
>>> from froide.helper import email_parsing
>>> import io
>>>
>>> data = b"To: 'John Doe [#123]' <john.doe@example.com>"
>>> parsed_email = email_parsing.parse_email(io.BytesIO(data))
>>> parsed_email.to
[('', "'John"), ('', ''), ('', '#123'), ('', ''), ("'", 'john.doe@example.com')]
```

PR #564, opened 2022-08-11
Filtering invalid email addresses

maxempoint:
Possible Fix for #515 and a test case for handling invalid email addresses.

I added a small function which checks if the email addresses that are returned by the email parser are valid. If not they are note appended in the address list.

Additionally, I added a test case with a new test mail including the incorrect mail address mentioned in the issue #515 .


froide/foirequest/tests/test_mail.py
```diff
@@ -296,6 +296,11 @@ def test_html_only_mail(self):
         self.assertIn(" ( https://", mail.body)
         self.assertIn("*peter.mueller@kreis-steinfurt.de*", mail.body)
 
+    def test_invalid_mail_address(self):
```
stefanw, line 299: Ideally this should be more a unit test (and less like a integration test): parsing a whole email just to test parsing of addresses makes this test not very specific. Also adding a large email with an image attachment (?!) is too much.

froide/helper/email_parsing.py
```diff
@@ -269,11 +269,21 @@ def try_decoding(encoded, encoding=None):
     return decoded
 
 
+def check_mail(email: str):
+    # Credit for regex to Anđela Niketić at https://stackabuse.com/python-validate-email-address-with-regular-expressions-regex/
+    regex_mail = r"([-!#-'*+/-9=?A-Z^-~]+(\.[-!#-'*+/-9=?A-Z^-~]+)*|\"([]!#-[^-~ \t]|(\\[\t -~]))+\")@([-!#-'*+/-9=?A-Z^-~]+(\.[-!#-'*+/-9=?A-Z^-~]+)*|\[[\t -Z^-~]*])"
```
stefanw, line 274: This regex is too complex to maintain for my taste (and still doesn't capture the complexity of proper email address parsing if you wanted to go there).
Please use the [Django email validator](https://docs.djangoproject.com/en/4.1/ref/validators/#validate-email) instead.

stefanw (changes requested):

froide/foirequest/tests/test_mail.py
maxempoint: Thank you for the feedback! I agree this is to unspecific. I will remove this test and do it more like this (without parsing a whole email):
```suggestion
    def test_invalid_mail_address(self):
        values = ["To: 'John Doe [#123]' <john.doe@example.com>"]
        ret = get_address_list(values)
        self.assertEqual('john.doe@example.com', ret[0].email)
```

froide/helper/email_parsing.py
maxempoint: Thank you also for this feedback! I used the Django Email validator as you suggested. I recognized that it also catched two cases were the emails in the test data were invalid (according to the django validator):

- In foirequest/tests/testdata/test_mail_05.txt on line 33: "CC: <KÃ¤mmerer.a@zdf.de>, <Biogiani.D@zdf.de>". The test `test_recipient_parsing` fails if the django email validator is used, because the first email-address is filtered. This seems so be a general problem with addresses which have umlaut characters in the local part. From what I found the django EmailValidator does [not support non-ASCII characters](https://code.djangoproject.com/ticket/27029) in the [local part](https://code.djangoproject.com/ticket/31053#comment:4). The [email-validator](https://pypi.org/project/email-validator/) for example has an `allow_smtputf8` option which seems to be not available in the django validator.
- In foirequest/tests/testdata/test_mail_08.txt on line 2: "X-Original-To: \<fragdenstaat class="de"\>\</fragdenstaat\>".

I could additionally change the mail addresses in those two cases in the test data if this makes sense? And would it make sense to use another validator because of the umlaut problem?

PR #568, opened 2022-08-16
Upgrade to Bootstrap 5 / introduce vite

stefanw:


frontend/javascript/snippets/bootstrap.ts
```diff
@@ -1,6 +1,17 @@
-import 'bootstrap/js/dist/alert'
+// FIXME: init consistently across components, some don't work yet
 import 'bootstrap/js/dist/collapse'
```
krmax44, line 2: It seems like initializing Bootstrap components is different for every one now. Tooltips for instance don't work when just globally imported. Also, the non-module files contain unnecessary code for jQuery users. Haven't found an elegant way yet to initialize them nicely.

stefanw: I don't think that is necessary! I had a problem there as well but it was because two bootstrap's were running on the page, the second undoing the actions of the first. You don't need to manually setup bootstrap components, it should work with the data-attribute APIs.

krmax44: Tooltips don't work at all for me when imported globally 🤔

stefanw: Another gotcha: I don't know why but the `froide` repo has the tendency to stick to its own old bootstrap 4.6. In `fragdenstaat_de` check version in `node_modules/froide/node_modules/bootstrap/package.json`

krmax44: nope, it's the correct version :/

pajowu:
In the status box, the label is now no longer full width, leading to the text breaking:

**old**:
Screenshot 2022-09-01 at 16 08 12


**this branch**:
Screenshot 2022-09-01 at 16 04 00

pajowu:
on the multi request page, there is an gradient element overlaying the entire page. this seems to be the `.filter-content::after` from the `pb-filter` component.

## Expected behaviour

The gradient is shown inside the public body filter only:

Screenshot 2022-09-01 at 16 41 42

## Actual behaviour

The gradient is shown at the bottom of the page, overlaying parts of the content and scrolling with the page. The public body filter has no gradient.

Screenshot 2022-09-01 at 16 35 45
Screenshot 2022-09-01 at 16 41 45

pajowu:
Description texts (at least on the make request page) are now shown next to the element instead of below it

## Previous behaviour


Screenshot 2022-09-01 at 16 48 07


## New behaviour
Screenshot 2022-09-01 at 16 47 07

pajowu:
The "Anfragetext prüfen" modal cannot be closed using the "Bearbeiten" button. The button just does nothing

pajowu:
If the "Anfragetext prüfen" modal is closed by clicking outside the modal, it (sometimes?) cannot be opened again.

Steps to reproduce
1. Write a new request
2. Click "Anfragetext prüfen"
3. Click outside of the dialog (on the gray area) -> the modal closes
4. Click "Anfragetext prüfen" again -> Nothing happens

pajowu:
The detail box of the foimessages now no longer has a two-column layout.

Screenshot 2022-09-01 at 17 36 29

krmax44:
the request message action buttons are now all centered, instead of them all being aligned at the end with only the comments button at start

image

PR #575, opened 2022-09-07
allow to set team for documents in admin

krmax44:


froide/document/admin.py
```diff
@@ -34,6 +35,11 @@ def execute_add_document_to_collection(admin, request, queryset, action_obj):
         CollectionDocument.objects.get_or_create(collection=action_obj, document=obj)
 
 
+def execute_set_team(admin, request, queryset, action_obj):
+    print(action_obj)
```
stefanw, line 39: No print calls, prefer logging.

stefanw (changes requested):
Apart from the print call, lgtm.

froide/document/admin.py
krmax44: argh, slipped through

PR #576, opened 2022-09-07
🛂 allow team members to view private documents

krmax44:
This PR touches on permissions, please review with care

### DIFF
--- a/froide/document/auth.py
+++ b/froide/document/auth.py
@@ -3,6 +3,8 @@ from django.utils.translation import override
 
 from crossdomainmedia import CrossDomainMediaAuth
 
+from froide.helper.auth import can_read_object
+
 
 class DocumentCrossDomainMediaAuth(CrossDomainMediaAuth):
     """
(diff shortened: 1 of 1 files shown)


froide/document/auth.py
```diff
@@ -25,6 +27,8 @@ def has_perm(self, request):
         obj = ctx["object"]
         if obj.user == request.user:
             return True
+        if has_authenticated_access(obj, request, "read"):
+            return True
         return super().has_perm(request)
```
stefanw, line 32: This should return the result of `can_read_object` directly and not call `super`. The [parent `has_perm`](https://github.com/stefanw/django-crossdomainmedia/blob/master/crossdomainmedia/auth.py#L88-L101) does the standard Django permission check which is already run as part of `can_read_object` so on deny it would unnecessarily run and fail twice.

stefanw (changes requested):

stefanw (approved):

PR #582, opened 2022-09-09
Fix public body proposal accepting

### LINKED ISSUE #581: Accepting change proposals for public bodies does not work
When moderating change proposals on public body entities, the selected proposal doesn't get applied.

According to the JS in https://github.com/okfde/froide/blob/main/froide/publicbody/static/js/publicbody_changes.js, the form fields should change when the moderator selects a proposal (radio input), but this isn't the case.
That leads to accepting proposals (and sending an email) without actually applying the proposed changes.

I noticed this behavior already by myself: I created a proposal on fragdenstaat.de to change a public body's fax number and mail address. Twenty minutes later, I got an email saying that my proposal was accepted. When looking at the public body's page on fragdenstaat.de, none of my proposed changes were applied.

After setting up a local instance of froide, I was able to reproduce this issue.

jasperroloff:
This PR aims to fix #581

It seems that the issue was caused by
1. having multiple `class` attributes on html elements and
2. by an issue where `'` was used instead of `"` to end an html attribute's value

This PR merges the multiple `class` attributes into single ones and replaces the `'` with `"`.

### DIFF
--- a/froide/publicbody/templates/publicbody/snippets/publicbody_proposals.html
+++ b/froide/publicbody/templates/publicbody/snippets/publicbody_proposals.html
@@ -29,7 +29,7 @@
       </td>
       <td>
         <div class="form-check">
-          <input class="form-check-input" type="radio" id="proposal_id' name="proposal_id" class="proposal" value="" required>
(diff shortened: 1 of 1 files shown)


stefanw:
Thanks!

PR #589, opened 2022-09-14
Mn make moderator badge optional

MagdaN:


froide/comments/models.py
```diff
@@ -1,5 +1,17 @@
+from django.db import models
+
 from django_comments.abstracts import CommentAbstractModel
+from django_comments.signals import comment_will_be_posted
 
 
 class FroideComment(CommentAbstractModel):
     ip_address = None
+    comment_as_normal_user = models.BooleanField(null=True)
```
stefanw, line 9: I would really like this turned around: `is_moderation` that can be `True`. This way the special handling is emphasized and not the 'normal' handling. Otherwise the question arises when reading: what does not normal mean?
Also I don't think `null=True` makes sense here, I can't see a difference between `False` and `None`.

froide/comments/templates/froide_comments/comment.html
```diff
@@ -20,7 +20,7 @@
         {# requester-/moderator-badges #}
         {% if comment.is_requester %}
           <span class="badge text-bg-blue-20 ms-1">{% trans "Requester" %}</span>
-        {% elif comment.is_moderator %}
+        {% elif comment.is_moderator and not comment.comment_as_normal_user %}
```
stefanw, line 23: Why not just `comment.is_moderation` (or whatever the Boolean field on the comment should be named)? The permissions of the user are no longer relevant when the comment is a moderation comment, right?

froide/comments/templates/froide_comments/form.html
```diff
@@ -22,10 +22,26 @@
 
   {% render_field form.comment %}
 
+  {% is_moderator as is_moderator %}
```
stefanw, line 25: The variable name should not shadow the template tag name to avoid confusion.

froide/foirequest/tests/test_web.py
```diff
@@ -534,7 +534,7 @@ def test_queries_foirequest_loggedin(self):
         - Problem reports - even for non-requester (+1)
         - ContentType + Comments for each FoiMessage (+2)
         """
-        TOTAL_EXPECTED_REQUESTS = 16
+        TOTAL_EXPECTED_REQUESTS = 18
```
stefanw, line 537: I'd really like to avoid adding queries to the request page. Especially because this looks like an 'n+1' query: an additional query per message, checking the the moderator status for every rendered comment form.

If an additional query really becomes necessary, please document it in the comment of the test.

froide/comments/templatetags/froide_comment_tags.py
```diff
@@ -13,3 +13,10 @@ def get_froide_comment_form(context, object):
         initial["name"] = request.user.get_full_name()
 
     return CommentForm(object, initial=initial)
+
+
+@register.simple_tag(takes_context=True)
+def is_moderator(context):
+    request = context["request"]
+    user = request.user
+    return user.groups.filter(permissions__codename="moderate").exists()
```
stefanw, line 22: Likely a query can be avoided (because the result will already be cached on the user) by using the [standard permission API of Django `user.has_perm`](https://docs.djangoproject.com/en/4.0/topics/auth/default/#permission-caching) instead of a custom ORM query.

froide/comments/templates/froide_comments/form.html
```diff
@@ -22,10 +22,26 @@
...
+  {% if is_moderator and not request.user == object.request.user %}
+  <div class="mb-3 row">
+    <div class="col-md-4 fw-bold">
+      {% trans "Moderator Badge" %}
+    </div>
+    <div class="col-md-8">
+      <div class="form-check">
```
stefanw, line 32: You can use the `{% render_field form.field_name %}` template tag from the `form_helper` template tag library.

stefanw: Also I just discovered that our comment forms don't use prefixes and ids clash because the form is embedded multiple times on the page. So labels are not properly linked to their fields. 😭

stefanw (changes requested):
Please turn the flag logic around, it will make understanding the code much easier.
When the comment knows if it's a moderation comment, the `is_moderator` annotation on the comment queryset can also be removed as it is no longer relevant.

froide/comments/templatetags/froide_comment_tags.py
stefanw: Also instead of a custom template tag, either use the existing `can_moderate_foirequest` (that has caching) or try `{{ perms.foirequest.moderate }}` (see [docs](https://docs.djangoproject.com/en/4.0/topics/auth/default/#permissions)).

froide/comments/models.py
MagdaN: I made it this way around to avoid extensive migrations, but can change it

froide/comments/templates/froide_comments/comment.html
MagdaN: If we turn the flag around, yes

froide/foirequest/tests/test_web.py
MagdaN: Yes, I was not happy with that, that's why I set this PR on draft

froide/comments/templatetags/froide_comment_tags.py
MagdaN: As far as I understood `can_moderate_foirequest ` and the `is_moderator` annotation on the comment don't behave the same. Which one should we use? If we use `can_moderate_foirequest ` staff users would also get the badge even if their are not in an moderation group (I think)

froide/comments/templates/froide_comments/form.html
MagdaN: Will try again

froide/comments/templates/froide_comments/form.html
MagdaN: sure

PR #591, opened 2022-09-16
Add is moderation flag in order to make moderator badge optional

MagdaN:


froide/comments/models.py
```diff
@@ -1,5 +1,17 @@
...
+from django_comments.signals import comment_will_be_posted
 
 
 class FroideComment(CommentAbstractModel):
     ip_address = None
+    is_moderation = models.BooleanField(default=False)
+
+
+def set_is_moderation_flag(comment, **kwargs):
+    is_moderation = kwargs.get("request").POST.get("is_moderation") == "on"
```
stefanw, line 13: This needs a server side check to see if the request is actually from a moderator.

froide/comments/forms.py
```diff
@@ -18,3 +18,16 @@ class CommentForm(DjangoCommentForm):
...
+        initial=True,
+        label=_("This is a moderation comment"),
+    )
+
+    def __init__(self, target_object, data=None, initial=None, **kwargs):
+        super().__init__(target_object, data=data, initial=initial, **kwargs)
+        object_id = self.target_object.id
+        for field in self.fields:
+            id = "id_{}_{}".format(field, object_id)
+            self.fields[field].widget.attrs["id"] = id
```
stefanw, line 33: Unsure if this is the way to go. Do [form prefixes](https://docs.djangoproject.com/en/4.0/ref/forms/api/#prefixes-for-forms) not work here? They would be cleaner but could be that they somehow don't work here.

stefanw (changes requested):

froide/comments/models.py
MagdaN: Added permission check

froide/comments/forms.py
MagdaN: Cool, did not know about form prefixes, tried ist, but prefixes change the name of the field and the [post view](https://github.com/django/django-contrib-comments/blob/master/django_comments/views/comments.py#L49) of django-comments can't deal with that.

stefanw (approved):

PR #593, opened 2022-09-20
✨ Add organization pages

pajowu:
Organization pages show the logo, name and description of organization as well as a list of its members (if their profile is public) with their name and profile description.

### DIFF
--- a/froide/helper/fields.py
+++ b/froide/helper/fields.py
@@ -8,6 +8,8 @@ from django.core.validators import (
 )
 from django.db import models
 
+from .widgets import ImageFileInput
+
 
 def validate_svg(f):
     # Find "start" word in file and get "tag" from there
@@ -36,6 +38,8 @@ class SVGAndImageFieldForm(forms.ImageField):
             allowed_extensions=get_available_image_extensions() + ["svg"]
         )
     ]
+
+    widget = ImageFileInput
 
     def to_python(self, data):
         try:
--- a/froide/organization/urls.py
+++ b/froide/organization/urls.py
@@ -1,0 +1,50 @@
+from django.urls import path
+from django.utils.translation import pgettext_lazy
+
+from .views import (
+    ChangeOrganizationMemberRoleView,
+    DeleteOrganizationMemberView,
+    InviteOrganizationMemberView,
+    JoinOrganizationView,
+    OrganizationDetailView,
+    OrganizationListView,
+    OrganizationUpdateView,
+)
+
+urlpatterns = [
+    path(
+        pgettext_lazy("url part", ""),
+        OrganizationListView.as_view(),
+        name="organization-list",
+    ),
+    path(
+        pgettext_lazy("url part", "<str:slug>/"),
+        OrganizationDetailView.as_view(),
+        name="organization-detail",
+    ),
+    path(
+        pgettext_lazy("url part", "<str:slug>/update/"),
+        OrganizationUpdateView.as_view(),
+        name="organization-update",
+    ),
+    path(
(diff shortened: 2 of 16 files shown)


froide/helper/fields.py
```diff
@@ -30,13 +30,24 @@ def validate_svg(f):
     return f
 
 
+class SVGAndImageFieldInput(forms.ClearableFileInput):
```
stefanw, line 33: There's nothing SVG specific about it right? We also have ProfilePhotoFileInput with an accompanying template in the codebase that is basically identical and can be replaced with this helper. Maybe this input widget should move to helper/widgets.py for consistency.

froide/organization/urls.py
```diff
@@ -0,0 +1,50 @@
...
+        OrganizationDetailView.as_view(),
+        name="organization-detail",
+    ),
+    path(
+        pgettext_lazy("url part", "<str:slug>/update/"),
+        OrganizationUpdateView.as_view(),
+        name="organization-update",
+    ),
+    path(
+        "organization/<str:slug>/invite/",
```
stefanw, line 31: Not translated!

froide/organization/urls.py
```diff
@@ -0,0 +1,50 @@
...
+        DeleteOrganizationMemberView.as_view(),
+        name="organization-delete_member",
+    ),
+    path(
+        pgettext_lazy("url part", "<str:slug>/change-member/<int:pk>"),
+        ChangeOrganizationMemberRoleView.as_view(),
+        name="organization-change_member",
+    ),
+    path(
+        "<str:slug>/join/<int:pk>/<str:secret>/",
```
stefanw, line 46: Not translated!

froide/organization/views.py
```diff
@@ -0,0 +1,208 @@
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        user_member = None
+
+        if self.request.user.is_authenticated:
+            try:
+                user_member = context["object"].organizationmembership_set.get(
+                    user=self.request.user
+                )
```
stefanw, line 37: Something like:
```
all_members = context["object"].organizationmembership_set.all()
user_member = all_members.filter(user=self.request.user).first()
```
avoids the var initialization and the try/except.

froide/organization/views.py
```diff
@@ -0,0 +1,208 @@
...
+        return OrganizationMembership.objects.filter(
+            organization__slug=self.kwargs["slug"],
+            status=OrganizationMembership.STATUS.INVITED,
+        )
+
+    def get_object(self):
+        member = super().get_object()
+        service = OrganizationService(member)
+        if not service.check_invite_secret(self.kwargs["secret"]):
+            raise Http404
```
stefanw, line 179: Maybe something friendlier? Like with `already_member`? Could be that a link was borked in an email.

stefanw (approved):

stefanw (changes requested):
There's some tiny things but there are two URL paths not marked for translations which should be fixed.

froide/organization/urls.py
pajowu: oops, good catch

pajowu: oops, good catch

froide/organization/views.py
pajowu: This would mean that we show a different response if the invite-id is invalid (404) vs if the secret is wrong (friendlier error message). Do we want to do that?

pajowu:
I fixed most of the review comment except the last one (see my comment in the conversation)

stefanw (approved):

PR #606, opened 2022-10-24
Only send out "Message sent"-emails after the message is marked as delivered

pajowu:
Note: This depends on the current delivery status checks. If we switch to the new, pygtails-based one, we have to re-work the code.


stefanw:
New mail delivery log parsing is deployed. 🥳 
Please adjust the tests and trigger the `FoiRequest.message_delivered` at an appropriate place.

stefanw:
I'm wondering if we need the extra model field (that would only be relevant for sent email messages).

Would it make sens to check delivery status in `save_delivery_status` and then send the confirmation email when it goes from "not sent" to `STATUS_SENT`?

PR #610, opened 2022-10-26
🐛 Use anchor href in guidance position calculation

pajowu:
The guidance position is calculated based on the FoiMessage.plaintext, but links are shortened when the message is rendered. Therefore we have to use the href property for calculating the guidance position


stefanw:
Closed in favor of #611

PR #621, opened 2022-11-23
Fix request search sorting

### LINKED ISSUE #620: Some sort options don't work
2 out of 5 sort options do not work.

Steps to reproduce:

- open a page with sort options (e. g. own requests or the overview page of a receiver with requests)
- click on [More search options] to display the combo box with the search options
- note the number of requests (should be > 0)
- select either "request date (newest first)" or "request date (oldest first)"
-> **expected behaviour:** the number of requests does _not_ change, requests are displayed
-> **actual behaviour:** the number of requests _does_ change to zero, no requests are displayed

I suggest to widen the sort option combo box to display the full sort option text (see screenshots below).

The following screenshots show the result for each of the 5 sort options. All sort options should result in the same number of requests found, but they don't.

# Sort option "default sort" works as expected

01 default sort

# Sort option "last message (newest first)" works as expected

02 last message (newest first)

# Sort option "last message (oldest first)" works as expected

03 last message (oldest first)

# Sort option "request date (newest first)" does not work

04 request date (newest first)

# Sort option "request date (oldest first)" does not work

05 request date (oldest first)

stefanw:
The changes to the request model (moving `first_message` to `created_at`) were not reflected in the search index. I would opt to keep the `created_at` timestamp out of public interfaces (might lead to confusion with later sent messages etc.)

I would propose to keep the search and the search index the same but load the actual first message's timestamp into the first_message date field on the search document.

Closes #620


froide/foirequest/documents.py
```diff
@@ -76,6 +76,12 @@ def prepare_public(self, obj):
     def prepare_campaign(self, obj):
         return obj.campaign_id
 
+    def prepare_first_message(self, obj):
+        return obj.first_message
+
+    def prepare_last_message(self, obj):
+        return obj.last_message
+
```
pajowu, line 84: Why a seperate function for `last_message` as well?

froide/foirequest/filters.py
```diff
@@ -228,8 +228,8 @@ class BaseFoiRequestFilterSet(BaseSearchFilterSet):
         widget=forms.HiddenInput(),
     )
 
-    created = django_filters.DateFromToRangeFilter(
-        method="filter_created",
+    first_message = django_filters.DateFromToRangeFilter(
+        method="filter_first_message",
```
pajowu, line 232: I would prefer to keep this as `first`, like we did for the last message filter

froide/foirequest/documents.py
stefanw: Ha, turns out you don't need this or `prepare_first_message`.

PR #624, opened 2022-11-24
Mn/add more autocomplete api endpoints

MagdaN:


froide/helper/widgets.py
```diff
@@ -97,6 +97,8 @@ def get_context(self, name, value, attrs):
                 ctx["values"] = [{"label": x, "value": x} for x in parse_tags(value)]
             else:
                 if self.model:
+                    if not type(value) is list:
```
stefanw, line 100: I'd prefer `isinstance(value, list)`.

PR #629, opened 2022-11-28
update foirequest settings

krmax44:
- 💄 use same headline level across settings
- 💄 card layout for foirequest settings ui
- 💬 update translations
- 🩹 keep project's team and request's team in sync:
    modifying a project's team will now update all corresponding requests to keep permissions in line


stefanw:
Ideally, translation changes with `make messagesde` to not include line numbers and only do `de` translations by default. Maybe Weblate will help with that eventually.

stefanw:
Yes, maybe that makes sense.

PR #630, opened 2022-11-29
Add new block to allow more actions in header

MagdaN:


froide/foirequest/templates/foirequest/header/header.html
```diff
@@ -55,6 +55,7 @@ <h4>{% trans "Special access" %}</h4>
             {% translate "Write a reminder to the public body!" %}
           </a>
         </li>
+        {% block aditional_actions %}{% endblock %}
```
krmax44, line 58: there's a typo, `aditional` needs an additional `d` :)

PR #642, opened 2022-12-16
✨ Link profile picture and name to profile

pajowu:
@krmax44 I'm not entirely sure about the styling. Right now the sender name is highlighted in blue, which is nice to make it clear that it is a link, but IMHO looks strange. What is your opinion on the styling of the link?

Screenshot of a message. The requester name is highlighted in blue


krmax44:
hmm, i think that might add to many clickable things in the message view. might be better to add this to the meta box or below the request title, like with projects? cleaning up the meta information display is on my todo list, i'd add the profile while working on that.

PR #648, opened 2022-12-21
✨ Improve API docs by using DRF-Spectacular

pajowu:


stefanw:
Due to our CSP settings this is currently not working in production because SwaggerUI and Redoc fail to load JS from CDNs. There are [settings to change the script URLs](https://drf-spectacular.readthedocs.io/en/latest/settings.html). The question is just how we get those onto our setup. Maybe just separate vite entrypoints?

pajowu:
DRF-Spectacular provides the sidecar package to serve these files locally. I will create a new PR to add those

Issue #649, opened 2022-12-27
Search leads to 404 page when switching from a high result page


politikundbildung:
Hi! I think I've found a bug while using FragDenStaat and hope this is the right place to report it.

To reproduce:

1. Search for a common term.
2. Jump to the last page of the results which will have an URL like https://fragdenstaat.de/behoerden/?page=398&q=schule&jurisdiction=&category=&classification=
3. In the sidebar on the left, click on a jurisdiction.

Result: You're redirected to the page https://fragdenstaat.de/behoerden/?page=398&q=schule&jurisdiction=nrw. This page doesn't exist, because the jurisdiction will naturally return less results than the whole country.

Expected result:
When choosing a jurisdiction from the sidebar, open the first page of results for that jurisdiction.

stefanw:
Thanks. The reason is that the `page` query parameter is reflected in the facetting links which does not make sense.

PR #651, opened 2023-01-04
Rework facetting code for list/search views

stefanw:


stefanw:
This would fix #649.

stefanw:
This would also fix #662 I believe.

PR #653, opened 2023-01-05
Add documents tab to foirequest page

stefanw:


pajowu (approved):

froide/foirequest/views/request.py
```diff
@@ -88,12 +94,43 @@ def get_context_data(self, **kwargs):
         return context
 
 
+MAX_DOC_COUNT = 100
```
pajowu, line 97: This is never used, is this on purpose?

PR #674, opened 2023-02-15
♻️ Replace exif-js with exifreade; fix exif support detection; cache image

pajowu:
This commit improves the handling of uploaded images in the document uploader of the attachment management component (i.e. the thing where you create a document out of uploaded images). It does three things:

1. It replaces the library we use for exif detection in the images. Until now we used exif-js which is no longer maintained and does not work in strict mode
2. It fixes the exif support feature detection
3. We now load the image in the component as a blob and stores it there. Until now the image was loaded by the img-tag in the browser and again by exif-js. However we can serve both from the component.

### DIFF
--- a/frontend/javascript/components/docupload/document-uploader.vue
+++ b/frontend/javascript/components/docupload/document-uploader.vue
@@ -202,16 +202,16 @@ export default {
     testExifSupport() {
       /*
       From Modernizr feature detection:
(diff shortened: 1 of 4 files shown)


krmax44 (changes requested):
Works great! It's possible to exclude parts of the module, maybe shrinks down the size? https://github.com/mattiasw/ExifReader#configure-a-custom-build

PR #707, opened 2023-08-21
♻️ rework slot exchange between django/vue

krmax44:
Before, every slot provided to a vue component via django html was stringified, then parsed again by the DOM. Now, a DOM copy of the `<template>` element is used instead (should improve performance).

To reduce confusion between Vue `<slot>`s and this, I propose to not use Vue syntax:

- Registering a template: `<template data-slot="foo" />` instead of `<template v-slot:foo />`
- Rendering the slot: `<django-slot name="foo" />` instead of `<slot name="foo">`

This also uses provide/inject instead of passing the slots as props, which has the additional benefit that `<django-slot name="foo">` can be used even in **nested components** without passing down the slot.

When using `createAppWithProps`, I suggest we always use the Vue component's name as the identifier:

```js
const els = document.querySelectorAll('document-uploader') // <document-uploader />

// instead of i.e. classes:
const els = document.querySelectorAll('.document-uploader') // <document-uploader class="document-uploader" />
```

### DIFF
--- a/froide/foirequest/templates/foirequest/attachment/manage.html
+++ b/froide/foirequest/templates/foirequest/attachment/manage.html
@@ -59,16 +59,16 @@
           <span class="visually-hidden">{% trans "Loading..." %}</span>
         </div>
       </div>
-      <template v-slot:convert-images>
+      <template data-slot="convert-images">
         <h3>{% translate "Convert images to documents" %}</h3>
       </template>
-      <template v-slot:documents>
+      <template data-slot="documents">
         <h3>{% translate "Documents" %}</h3>
         <p>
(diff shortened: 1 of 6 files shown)


stefanw (approved):
I like. Single annoyance is the additional `<div>` wrapper around slot content. Could this be a `<template v-if="true" ref="container">`? ([`v-if` in order to prevent rendering as native element](https://vuejs.org/api/built-in-special-elements.html#template))

krmax44:
@stefanw the `<div>` is just a placeholder and should be replaced with the slot content (not wrapping it)

krmax44:
It seems to me that there are no changes required in the other repos, as we apparently don't pass slots from Django to Vue there. Someone double-check though 😉

PR #708, opened 2023-08-21
🚧 error page redesign

krmax44:


frontend/styles/variables.scss
```diff
@@ -87,7 +87,8 @@ $spacers: (
 // Define the maximum width of `.container` for different screen sizes.
 $container-max-widths: (
   sm: 540px,
-  md: 780px, // changed from original Bootstrap 720px for extra space!
+  md: 780px,
+  // changed from original Bootstrap 720px for extra space!
```
pajowu, line 91: Shouldn't this go above the md-definition?

frontend/javascript/snippets/search.ts
```diff
@@ -1,48 +1,25 @@
...
-  if (event.target == null) {
-    return
-  }
-  if (event.target instanceof HTMLElement) {
-    const focussable = event.target.querySelector('[data-focus]')
-    if (focussable instanceof HTMLInputElement) {
-      focussable.focus()
+    if (bestMatch !== undefined) {
+      form.action = bestMatch
+      select.value = bestMatch
```
pajowu, line 23: whew, interesting logic :D but seems to work well

frontend/styles/variables.scss
krmax44: Seems like my linter moved it down from the end of the line.

PR #711, opened 2023-09-18
🐛 Set confirmation_sent to true for old messaged in db

pajowu:


stefanw:
I don't understand the change here? This changes an already applied migration to Add

pajowu:
This migrations is meant to run when the new mail log parser is installed. We need to set `confirmation_sent` for all previous sent messages, the default was mistakenly set to `False`, so that we sent out the "Message sent"-E-Mails again. By first adding it with a default of `True`, we set it for all the previous messages and then change the default in the `AlterField` step

stefanw:
Ah, now I get it the meaning of the two operations. However, this change will not affect any system that already applied 0064 (ie. our production system) as the migration will not run again – the migration system does not detect the change in 0064.

pajowu:
Yes, i’m aware of that. On our production system we already chose another path (setting the flag manually for the old messages), this is meant for other systems migrating now

stefanw (approved):

fin:
speaking for the other systems: thank you! 🎉

PR #714, opened 2023-10-05
✨ Allow users to redact description

pajowu:


froide/foirequest/forms/request.py
```diff
@@ -449,3 +450,43 @@ def save(self) -> List[str]:
...
+        val = self.cleaned_data["description"]
+        if not val:
+            return []
+        try:
+            val = [int(x) for x in val.split(",")]
+        except ValueError:
+            raise forms.ValidationError("Bad value")
+        return val
+
+    def save(self, request):
```
stefanw, line 469: This is a `FoiRequest` and should be named `foirequest` to avoid confusion with `HttpRequest`

froide/foirequest/forms/request.py
```diff
@@ -449,3 +450,43 @@ def save(self) -> List[str]:
...
+        )
+        first_message = request.first_outgoing_message
+        if request.description_redacted in first_message.plaintext_redacted:
+            first_message.plaintext_redacted = first_message.plaintext_redacted.replace(
+                request.description_redacted, redacted_description
+            )
+            first_message.save()
+        else:
+            messages.add_message(
+                request,
```
stefanw, line 483: This needs to be a `HttpRequest` not a `FoiRequest`. Possibly return a value and handle 'flash messages' in the view instead?

froide/foirequest/templates/foirequest/header/header.html
```diff
@@ -103,82 +92,104 @@ <h2 class="request-title">{{ object.title }}</h2>
...
+          </div>
         {% endif %}
 
         {# description #}
         <div class="mt-3">
           <div class="request-descr request-descr--collapsed">
-            {{ object.get_description|urlizetrunc:40|linebreaks }}
+
+            {% redact_request_description object request as redacted_description %}
+            {{ redacted_description|urlizetrunc:40|linebreaks }}
```
stefanw, line 115: `redact_description` is already 'urlized' in `mark_redacted`, so this would apply `urlizetrunc` twice.

froide/foirequest/templatetags/foirequest_tags.py
```diff
@@ -151,6 +153,22 @@ def redact_message(message, request):
     return content
 
 
+@register.simple_tag
+def redact_request_description(foirequest, request):
```
stefanw, line 157: I would suggest adding return types here e.g. either `str` or `SafeString` so it's clear if the string has been escaped to avoid double escaping when the code is reused.

stefanw (changes requested):
Found two bugs, looks good otherwise. Glad we are fixing this!

PR #718, opened 2023-10-09
🔨 add djlint to pre-commit

krmax44:


pajowu:
Does this change any of the existing files?

PR #719, opened 2023-10-12
🩹 fix select auto-submit on request search

krmax44:
was deleted accidentally in a63f8be, brought back as a simpler utility


pajowu (approved):

stefanw:
Just FYI, there's another `autosubmit` utility that auto-submits forms, but is only loaded on one specific page: https://github.com/okfde/froide/blob/main/froide/account/static/account/go.js

PR #720, opened 2023-11-08
✨ Create problem report for bounced delivery reports

pajowu:


froide/foirequest/signals.py
```diff
@@ -446,6 +447,18 @@ def save_delivery_status(
 
     if status == "sent":
         send_foimessage_sent_confirmation(message)
+    elif delivery_status.is_failed():
+        try:
+            ProblemReport.objects.get(
```
stefanw, line 452: Maybe if condition and `.filter(...).exists()` because we don't need the object.

stefanw (approved):

PR #722, opened 2023-11-16
Add admin action and task to unzip FoiAttachment

stefanw:
It's a handy feature, but the following conceptual problems are unresolved:

- do new attachments keep track of their source attachment? The `converted` FK (points to converted file) and `is_converted` (is `True` on original) flag are intended to work the other way around. Changing the meaning just for this feature doesn't work well with other checks on `converted`. Maybe we need to refactor the meaning everywhere?
- do we trigger the unpacking automatically? I'm a bit afraid of recursive Zip files are zip bombs. Right now there is an admin action to trigger it manually and task is killed after three minutes.

### DIFF
--- a/froide/foirequest/admin.py
+++ b/froide/foirequest/admin.py
@@ -756,6 +756,7 @@ class FoiAttachmentAdmin(admin.ModelAdmin):
         "convert",
         "ocr_attachment",
         "make_document",
+        "unpack_zipfile",
     ]
 
     def get_queryset(self, request):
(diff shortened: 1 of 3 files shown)


froide/foirequest/admin.py
```diff
@@ -824,6 +825,13 @@ def ocr_attachment(self, request, queryset):
         for att in queryset:
             ocr_pdf_attachment(att)
 
+    @admin.action(description=_("Unpack ZIP file"))
+    def unpack_zipfile(self, request, queryset):
+        from .tasks import unpack_zipfile_attachment_task
+
+        for att in queryset:
+            unpack_zipfile_attachment_task(att)
```
pajowu, line 833: Doesn't this run the task inline instead of as a task?

froide/foirequest/utils.py
```diff
@@ -880,3 +885,52 @@ def select_foirequest_template(foirequest, base_template: str):
 
     templates.append(base_template)
     return templates
+
+
+ZIP_BLOCK_LIST = set(["__MACOSX"])
```
pajowu, line 890:
```suggestion
ZIP_BLOCK_LIST = set(["__MACOSX", ".DS_Store"])
```

PR #756, opened 2024-04-22
💚 Use pre-commit in ci

pajowu:


krmax44:
We're using pre-commit in CI now.

PR #760, opened 2024-05-04
refactor: minor alterations

tyliec:


pajowu:
Good catches, I'll make some small changes and then merge it

PR #769, opened 2024-05-13
css clean-up, breadcrumbs

krmax44:
upstream changes for okfde/fragdenstaat_de#494

breadcrumbs are now more flexible, type-safe and better documented - simply extend views with the provided class:
https://github.com/okfde/froide/blob/815abaf8308a81a6428f98961f3cd305e6037219/froide/helper/breadcrumbs.py#L23-L42


froide/helper/breadcrumbs.py
```diff
@@ -0,0 +1,124 @@
...
+
+    class MyView(BreadcrumbView):
+        def get_breadcrumbs():
+            return [("Foo", "/foo/"), "Bar"]
+    """
+
+    breadcrumbs: Optional[Sequence]
+
+    def get_breadcrumbs(self) -> Union[BreadcrumbItems, Breadcrumbs]:
+        raise NotImplementedError("No breadcrumb provider implemented")
```
pajowu, line 46: To simplify the interface the the view, we could let `get_breadcrumbs` return `breadcrumbs` if set and raise otherwise, similar to what django does with get_queryset on generic views [here](https://github.com/django/django/blob/86e13843c2ab510fba1de84588efe7fd03555531/django/views/generic/detail.py#L61)

froide/helper/breadcrumbs.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@dataclass
+class BreadcrumbItem:
+    # will be displayed as the link text
+    title: str
```
pajowu, line 52: Is `str` the right type for (lazy) translatable strings? https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#lazy-translations

froide/helper/breadcrumbs.py
```diff
@@ -0,0 +1,124 @@
...
+            return Breadcrumbs(items=items)
+
+
+def normalize_breadcrumb(breadcrumb: Union[str, BreadcrumbTuple]) -> BreadcrumbTuple:
+    if type(breadcrumb) is tuple:
+        return breadcrumb
+    elif type(breadcrumb) is str:
+        return (breadcrumb, None)
+
+    logger.error("Received breadcrumb that is neither a tuple nor a string", breadcrumb)
```
pajowu, line 113: Why not throw an exception here and in L121?

froide/helper/breadcrumbs.py
```diff
@@ -0,0 +1,124 @@
...
+    You can also directly return a `Breadcrumbs` instance.
+
+    Example implementation:
+
+    class MyView(BreadcrumbView):
+        def get_breadcrumbs():
+            return [("Foo", "/foo/"), "Bar"]
+    """
+
+    breadcrumbs: Optional[Sequence]
```
pajowu, line 43: Sequence of what? Can this be typed?

froide/helper/breadcrumbs.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __len__(self):
+        return len(self.items)
+
+    def __add__(self, items: BreadcrumbItems):
+        """
+        Add items using the plus operator: breadcrumbs + [...]
+        """
+        self.items += items
+        return self
```
pajowu, line 85: Nice!

froide/helper/templatetags/breadcrumb_helper.py
```diff
@@ -1,31 +1,10 @@
...
-        return view.get_breadcrumbs(context)
-
-    if hasattr(view, "breadcrumbs"):
-        return map(normalize_breadcrumb, view.breadcrumbs)
-
-
-@register.filter
-def has_link(value):
-    return type(value) == tuple and len(value) == 2
+    return Breadcrumbs.from_view(view, context)
```
pajowu, line 10: `from_view` expects a non-null view, should we remove support vor `get_breadcrumbs` with a `None`-view?

pajowu: Or add

```python
if view is None:
    return None
```

pajowu (approved):
LGTM, some small thoughts in the comments. But can be merged either way IMHO

froide/helper/templatetags/breadcrumb_helper.py
krmax44: Not sure what you mean. `from_view` is able to handle `view = None`, and returns `None` in that case

froide/helper/breadcrumbs.py
krmax44: I should probably not mutate it, though…

froide/helper/breadcrumbs.py
krmax44: I think `BreadcrumbItem` shouldn't deal with Promises. The `Breadcrumbs` class calls `str` before passing the title to `BreadcrumbItem`

PR #774, opened 2024-06-03
Generate PDFs and ZIP packages for messages

stefanw:
- Add individual PDFs per message to foirequest package download
- Add download options to message meta container
- Some cleanup

### DIFF
--- a/froide/foirequest/templates/foirequest/body/message/meta_container.html
+++ b/froide/foirequest/templates/foirequest/body/message/meta_container.html
@@ -243,13 +243,23 @@
                             </a>
                         </li>
                     {% endif %}
-                    {% if message.can_get_original_from_imap and object|can_write_foirequest:request %}
-                        <li>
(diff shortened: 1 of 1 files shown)


froide/foirequest/templates/foirequest/pdf/_foimessage.html
```diff
@@ -0,0 +1,42 @@
+{% load i18n %}
+<section>
+    <h2>{{ message.timestamp|date:"DATETIME_FORMAT" }} {{ message.real_sender }}</h2>
+    <dl>
```
krmax44, line 4: Maybe make this a table for easier readability?

krmax44 (approved):

froide/foirequest/templates/foirequest/pdf/_foimessage.html
stefanw: This gets rendered by weasyprint to PDF (no browser, no other CSS). The idea was to make it roughly look like Outlook‘s print style maybe. Not sure if table helps here.

PR #775, opened 2024-06-04
Add reason to publicbody proposal / change form

stefanw:


krmax44:
Should we add a reason field also to the change proposal?

PR #797, opened 2024-07-10
fix image document page sortability

bikubi:
vuedraggable, broken with Vue3, replaced with vueuse's sortable (thx @krmax44 for the recommendation!)


frontend/javascript/components/docupload/image-document-pages-sortable.vue
```diff
@@ -0,0 +1,51 @@
...
+which looks like the most reasonable replacement for vuedraggable (Vue2 only);
+other alternatives did not work (vuedraggable-next and a different thin SortableJS wrapper). */
+
+import { useSortable } from '@vueuse/integrations/useSortable'
+import { ref, defineEmits, computed } from 'vue'
+import ImagePage from './image-page.vue'
+
+const props = defineProps({ pages: Array })
+
+// once updated to Vue>=3.4
```
krmax44, line 12: we could also upgrade right now! :)

frontend/javascript/components/docupload/image-document-pages-sortable.vue
```diff
@@ -0,0 +1,51 @@
...
+
+<style lang="scss" scoped>
+.pages {
+  display: flex;
+  flex-wrap: nowrap;
+  align-items: baseline;
+  overflow: auto;
+  overflow-x: scroll;
+  overflow-scrolling: touch;
+  padding-bottom: 2rem;
```
krmax44, line 49: I prefer using utility classes for spacing, whenever possible, for consistency (i.e. `pb-3`). Maybe also `d-flex flex-nowrap align-items-baseline`? Stops the CSS from growing.

frontend/javascript/components/docupload/image-document.vue
```diff
@@ -34,19 +34,17 @@
...
-            @end="drag = false">
-            <image-page
-              v-for="page in pages"
-              :key="page.pageNum"
-              :page="page"
-              :page-count="pages.length"
-              @pageupdated="$emit('pageupdated', { document, ...$event })"
-              @splitpages="splitPages" />
-          </draggable>
+          <!-- akward @update because Vue 2-ish vs 3.2 interop -->
```
krmax44, line 37: Upgrading this to vue 3 syntax probably requires more changes downstream, right? 😬

frontend/javascript/components/docupload/image-document-pages-sortable.vue
bikubi: I can give it a quick try - I've glanced at 3.2→3.3→3.4 changelogs/deprecations and they seemed fairly benign. But I feel like I would have to test *all* Vue code to avoid subtle breakage like the situation here...

frontend/javascript/components/docupload/image-document.vue
bikubi: This awkwardness happens where Vue2/3 (resp.: options/composition API based) components touch. *Upgrading* our way out of the awkwardness means upgrading the parent component to Vue3/Composition. Which would then shift the 2/3 touching points to other components.. end cascade Vue3 everywhere :)
...I think.
I'm not familiar with mixed Vue2/3 codebases. I gave a mixed/hybrid SFC a try but quickly ran into weird errors and unsupported dealbreakers. It might work in other places. Not sure if/how/when to *upgrade* stuff to Vue3, other than on a "is it new / is it broken now" basis.

PR #813, opened 2024-08-12
wip: postupload

krmax44:
todos:

- [ ] after merge: add translation messages
- [ ] after deploy: add scheduled task to delete stale drafts


froide/foirequest/forms/postal.py
```diff
@@ -142,3 +151,73 @@ def create_attachment(self, message, upload):
...
+
+
+class ThinModelChoiceIterator(ModelChoiceIterator):
+    def __init__(self, field):
+        self.field = field
+        # basically empty the queryset so the thousands of choices are not rendered into context
+        self.queryset = field.queryset[:0]
+
+
+class ThinModelChoiceField(ModelChoiceField):
```
krmax44, line 163: I don't quite understand what this does

froide/foirequest/forms/postal.py
```diff
@@ -142,3 +151,73 @@ def create_attachment(self, message, upload):
...
+    iterator = ThinModelChoiceIterator
+
+
+class PostalEditForm(MessageEditMixin, JSONMixin, forms.Form):
+    sent = forms.TypedChoiceField(
+        choices=(
+            (0, _("I have received the letter")),
+            (1, _("I have sent the letter")),
+        ),
+        coerce=lambda x: bool(int(x)),
```
krmax44, line 173: Is that safe?

froide/foirequest/urls/request_urls.py
```diff
@@ -105,6 +107,16 @@
         upload_postal_message,
         name="foirequest-upload_postal_message",
     ),
+    path(
+        pgettext_lazy("url part", "<slug:slug>/upload-postal-message-create/"),
```
krmax44, line 111: rename and add i18n later

frontend/javascript/components/docupload/file-document.vue
```diff
@@ -183,15 +224,25 @@ import { DocumentMixin } from './lib/document_utils'
...
   mixins: [I18nMixin, DocumentMixin],
-  props: ['config', 'document'],
+  props: [
+    'config',
+    'document',
+    'hideSelection',
+    'hideAdvancedOperations',
+    'iconStyle',
+    'showBasicOperations',
+    'highlightRedaction'
```
krmax44, line 244: do we need these props to support both the old and the new flow simultaneously? looks like these are very specific props that might overcomplicate the component

frontend/javascript/components/docupload/file-document.vue
```diff
@@ -56,6 +77,15 @@
           <span class="visually-hidden">{{ i18n.openAttachmentPage }}</span>
         </a>
 
+        <div
+          v-if="highlightRedaction"
+          class="redaction"
+          :class="{ 'redaction__is-redacted': document.is_redacted }">
+          <div class="redaction-badge">
```
krmax44, line 84: better to just use a regular bootstrap badge, like `badge text-bg-success` or `badge text-bg-warning`

frontend/javascript/components/postupload/app-shell.vue
```diff
@@ -0,0 +1,340 @@
+<script setup>
+import { ref, computed } from 'vue'
+import OnlineHelp from './online-help.vue'
+
+const props = defineProps({
+  step: Number,
+  steps: Array
+})
+
+const progressDesktop = computed(() => props.step / (props.steps.length - 1))
```
krmax44, line 10: maybe a bit clearer: this one starts from zero, the other one from one

frontend/javascript/components/postupload/app-shell.vue
```diff
@@ -0,0 +1,340 @@
...
+
+<!-- unscoped!
+  well, scoped to the current django route .../post-hochladen/ since the global css is not ejected+collected
+  else would have to guard via body.appshell
+-->
+<style lang="scss">
+// TODO: arbitrarily switch to "appshell" under this breakpoint
+//   not really usable, but also not broken. WIP
+// TODO: sync this with PostUpload's matchMedia and scss
+$breakpoint: 1000px;
```
krmax44, line 94: better use the bootstrap provided layout, grid system, utilities and variables for consistency and maintainability

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -0,0 +1,1494 @@
...
+const formStatusChoices = computed(() =>
+  props.status_form.fields.resolution.choices.filter((choice) => {
+    if (formIsSent.value) {
+      switch (choice.value) {
+        case 'successful':
+        case 'partially_successful':
+        case 'not_held':
+        case 'refused':
+          return false
+      }
```
krmax44, line 104:
```suggestion
      const badCombinations = ['successful', 'partially_successful', 'not_held', 'refused']
      if (badCombinations.includes(choice.value)) return false
```

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -0,0 +1,1494 @@
...
+        console.log('delay done')
+        resolve(x)
+      }, duration)
+    })
+  }
+}
+*/
+
+// this could possibly move to lib/vue-helper.ts
+const vBsTooltip = {
```
krmax44, line 37: great idea! could also be a helper component

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -0,0 +1,1494 @@
...
+  const rect = header.getBoundingClientRect()
+  if (rect.bottom < 0) {
+    // timeout to make it feel less jarring
+    setTimeout(() => header.scrollIntoView({ behavior: 'smooth' }), 500)
+  }
+}
+
+const mobileHeaderTitle2 = computed(() => {
+  switch (step.value) {
+    case 1100:
```
krmax44, line 200: where do the numbers come from?

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -0,0 +1,1494 @@
...
+const progressStep = computed(() =>
+  // map to {0, 1, 2}
+  Math.min(Math.floor((step.value - 1000) / 1000), 2)
+)
+
+const gotoStep = (nextStep) => {
+  stepHistory.push(nextStep || getNextStep())
+}
+
+const gotoValid = computed(() => {
```
krmax44, line 234: maybe we should use a proper state machine for this, like xstate? this is very confusing to me

froide/foirequest/forms/postal.py
bikubi: When I serialized the Form with it's `publicbody` field in the template to pass it via prop to vue, it serialized *all* publicbodies into 4MB of JSON (due to `queryset=PublicBody.objects.all()`). Somewhere I found this technique to override the `iterator` to the field so it still works for `save()` and such but doesn't serialize all objects... this is probably not idiomatic or even clean, but the best I could come up with :|

froide/foirequest/forms/postal.py
bikubi: unsafe in what way? (I just copied it from elsewhere in the code.)

frontend/javascript/components/docupload/file-document.vue
bikubi: agree, i will address this in the next pass, essentially replacing the "old style" components

froide/helper/form_utils.py
```diff
@@ -18,9 +19,35 @@ def get_data(error):
...
+# FIXME WIP this is maybe not the smartest way to do this
+# use simplejson instead?
+def serialize_extra(obj):
+    if isinstance(obj, Decimal):
+        return {
+            "__Decimal": True,
+            "intValue": int(obj * 100),
+            "strValue": str(obj),
+        }
+    if obj.__class__.__name__ == "PublicBody":
```
pajowu, line 31:
```suggestion
    if isinstance(obj, PublicBody):
```

froide/helper/form_utils.py
```diff
@@ -18,9 +19,35 @@ def get_data(error):
     return error.get_json_data()
 
 
+# FIXME WIP this is maybe not the smartest way to do this
+# use simplejson instead?
+def serialize_extra(obj):
+    if isinstance(obj, Decimal):
+        return {
+            "__Decimal": True,
```
pajowu, line 27: is this a nice way or should we add a tag field that specifies the type of the object serialized here? e.g. something like

```suggestion
            "__type__": "__Decimal",
```

frontend/javascript/components/postupload/app-shell.vue
bikubi: this one is complex :)
- appshell needs to hide `#header`, but header from the base template can't have class `d-lg-none` globally, only when appshell is used...
- by default, BS doesn't have responsive `position-*-fixed` classes. we could [enable them](https://getbootstrap.com/docs/5.3/utilities/api/#enable-responsive)
- but I do more in the `@media` blocks, like `top`, `bottom` (which could be worked around, I guess)
- before I found out about `position-foo-fixed` I implemented one global scss+css var in froide/frontend/styles/variables: `$froide-appshell-breakpoint` which equals BS's `lg` -- because IMO it makes sense to have it defined centrally/DRY. It wouldn't make sense to change *one* of the `d-lg-foo` in appshell; they need to be in sync. This adheres to BS's grid breakpoints without suggesting that individual `d-lg-foo` classes can be adjusted individually. (If we go with this solution I'll still have to fix the logic; I used `max-width` while BS's breakpoints all cascade `min-width`ly...)

frontend/javascript/components/postupload/post-upload.vue
bikubi: For the WIP I've used the autogenerated frame IDs/numbers from Figma, ähem. Replacing them with something better would go hand-in-hand with a proper state machine :) (but will require some reworking of the step-ID-to-file-index arithmetic...)

frontend/javascript/components/postupload/post-upload.vue
bikubi: yes! having a visual representation will be very beneficial

froide/helper/form_utils.py
bikubi: i like your suggestion; tbh i don't use the type info anywhere - i just felt dirty not doing anything :) yagni?

froide/foirequest/views/message.py
```diff
@@ -151,6 +157,311 @@ def add_postal_message(request, slug):
...
+
+            messages.add_message(request, messages.SUCCESS, success_message)
+            url = reverse(
+                # after a successfully saved form, redirect to request
+                "foirequest-show",
+                kwargs={"slug": foirequest.slug},
+            )
+            return redirect(url)
+        else:
+            if is_fetch(request):
```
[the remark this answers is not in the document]
bikubi: yes. resolved by 998ad92a8

froide/foirequest/views/message.py
```diff
@@ -151,6 +157,311 @@ def add_postal_message(request, slug):
     )
 
 
+@allow_write_foirequest
+def upload_postal_message_create(request, foirequest):
+    message = FoiMessage(request=foirequest, kind=MessageKind.POST)
+    # hack, prevent: null value in column "timestamp" of relation "foirequest_foimessage" violates not-null constraint
+    # this will be set to "proper" (noon) when finally submitted
+    message.timestamp = datetime.datetime.now()
```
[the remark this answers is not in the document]
bikubi: oof, hard to tell (for me). i think it *shouldn't*, because, as stated in the comment, it's a temporary non-null replacement, that shouldn't exist for too long.
but speaking of timezone problems, `registered_mail_date` from your other comment might cause problems? it's supposed to be a "timeless" date, so it gets set to midnight. i found `date` field logic hard to understand...

froide/foirequest/views/message.py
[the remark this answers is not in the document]
bikubi: whoops, unresolved -- a normal `form.submit` sends `Accept: ...,*/*` so it does `accepts("application/json")`, so this is no way to distinguish from a `fetch`. I'll revert this unless you know a better solution

frontend/javascript/components/postupload/post-upload.vue
bikubi: plan of attack:
1. attach the current implementation as-is to history api (like ` .../#1100'`)
2. see if xstate can a. handle our state machine part and b. work with hash-history
3. if not, consider a lightweight, handrolled state machine implementation

frontend/javascript/components/postupload/post-upload.vue
krmax44: @krmax44

froide/foirequest/views/message.py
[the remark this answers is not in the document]
bikubi: @krmax44 suggest adding a constraint that allows this field to be nullable as long is_draft UNLESS this is the only timestamp/created_at-ish info

froide/foirequest/views/message.py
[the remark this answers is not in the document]
bikubi: @krmax44 will look into turning this into a proper api endpoint

froide/foirequest/forms/postal.py
bikubi: @krmax44 I've just looked into this again
- `raw_id_fields` and ForeignKeyRawIdWidget only work for Admin, as far as I can tell
- is using an integer, like [here](https://github.com/okfde/froide/blob/main/froide/document/documents.py#L29) a similar idea? seems hacky
- I suspect this will be moot anyway if/when we turn this into an API?

krmax44: how about a [hidden input widget](https://docs.djangoproject.com/en/5.1/ref/forms/widgets/#hiddeninput)? the linked example (point 2) is about a elasticsearch model, which looks the same but is quite different 🙃

froide/settings.py
```diff
@@ -566,6 +566,8 @@ def is_pkce_required(client_id):
             "about": "/about/",
             "help": "/help/",
             "throttled": "/help/",
+            # TODO english?
+            "help_postupload_redaction": "/hilfe/plain/funktionen-der-plattform/schwaerzungen-durchfuehren/",
```
krmax44, line 570: this should go into the theme repo: https://github.com/okfde/fragdenstaat_de/blob/40c44c2188808c5feb1d49dbfde0093b34e435e8/fragdenstaat_de/settings/base.py#L719

frontend/javascript/components/postupload/post-upload.vue
krmax44: maybe vue-router could also work for this

frontend/javascript/components/postupload/post-upload.vue
krmax44, line 1: todo: split this up into smaller views; move the larger logic to a javascript module/class

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -0,0 +1,1663 @@
...
+    cursor: pointer;
+    text-decoration: underline;
+    color: var(--bs-link-color);
+    /* padding-left matches bootstrap's form-group>input for indent */
+    padding: 0.25rem 0.25rem 0.25rem 0.75rem;
+  }
+}
+
+/* bootstrap doesn't provide this responsive position class out of the box
+ * (it could be compiled in with utils/helpers)
```
krmax44, line 1655: this can be configured: https://getbootstrap.com/docs/5.3/utilities/api/#using-the-api

froide/settings.py
bikubi: it's there: https://github.com/okfde/fragdenstaat_de/blob/bikubi/postupload-wip3/fragdenstaat_de/settings/base.py#L726

but since post-upload.vue is part of froide, and this help url is kind of expected (short of hard coded), isn't it weird to leave it undefined, then? or should the text which contains this link be also moved to the theme repo?

frontend/javascript/components/postupload/post-upload.vue
bikubi: but doesn't vue-router also always expose the state to the back-button, leaving states reachable which shouldn't be? resp., necessitate complex navigation guards, which are state-machine-like...

frontend/javascript/components/postupload/post-upload.vue
bikubi: done! wasn't sure if this one case was worth the overhead, but also hadn't seen that we already have utilities configured.. thx

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -0,0 +1,1685 @@
+<script setup>
+import { ref, reactive, computed, defineProps, nextTick } from 'vue'
+import SimpleStepper from './simple-stepper.vue'
+// import PublicbodyChooser from '../publicbody/publicbody-chooser'
+import PublicbodyChooser from '../publicbody/publicbody-beta-chooser'
+// TODO linter wrong? the two above are just fine...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import DocumentUploader from '../docupload/document-uploader.vue'
```
krmax44, line 8: I guess this is because in line 190, you define `const documentUploader = ref()`. This introduces ambiguity, since component names are case-insensitive. While Vue uses the correct one (the imported component) in rendering, it seems ESLint doesn't like it so much.

PR #828, opened 2024-09-09
🐛 Redact names with spaces if split in attachment name

### LINKED ISSUE #827: Censor information in files names.

pajowu:
Closes #827


stefanw (approved):
Only concern is for pathological cases where the name has many spaces or can override the file extension (already possible).

PR #868, opened 2024-10-17
Post-upload i18n + l10n DE

bikubi:
- Use the "classic" Django→Vue `i18n.foo` implementation, postpone vue3-gettext due to build complexity (this work is not moot, it is still valuable for vue3-gettext later)
- I've added all DE translations manually, because locally `make messagesde` produced a horrible mess. Now I understand the weblate integration better. I think/hope that weblate will pick up the .po file just fine.


frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -1683,3 +1677,12 @@ addEventListener('hashchange', () => {
 }
 
 </style>
+
+<style lang="scss">
+/* make links stand out more in alert boxes.
+   needs to be unscoped because i18n/v-html */
+
```
pajowu, line 1684: In that case it should go into `main.scss` or similar, not in the vue-component

frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -1682,4 +1676,9 @@ addEventListener('hashchange', () => {
   }
 }
 
+/* make links stand out more in alert boxes. needs :deep because <a> in gettext msg */
```
krmax44, line 1679: Better to use `a.alert-link`: https://getbootstrap.com/docs/5.3/components/alerts/#link-color

PR #869, opened 2024-10-21
staff auth decorator, form labels

krmax44:
required for https://github.com/okfde/fragdenstaat_de/pull/601

### DIFF
--- a/froide/helper/auth.py
+++ b/froide/helper/auth.py
@@ -2,6 +2,7 @@ from functools import lru_cache, reduce
 from operator import or_
 
 from django.contrib.auth import get_permission_codename
+from django.core.exceptions import PermissionDenied
 from django.db.models import Q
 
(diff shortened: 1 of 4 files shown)


froide/helper/auth.py
```diff
@@ -224,6 +225,15 @@ def get_user_filter(request, teams=None, fk_path=None):
     return filter_arg
 
 
+def require_staff(view_func):
+    def decorator(request, *args, **kwargs):
+        if not hasattr(request, "user") or not request.user.is_staff:
+            raise PermissionDenied
```
pajowu, line 231: Shouldn't `request.user` always be set? https://docs.djangoproject.com/en/4.2/ref/request-response/#attributes-set-by-middleware

pajowu (approved):
One small question, but can be merged IMO

PR #876, opened 2024-10-25
postupload fixes

bikubi:
fcf2fdb855d94ac60a35b02eea83669d3b8168b7 is the meat here

### DIFF
--- a/frontend/styles/maps.scss
+++ b/frontend/styles/maps.scss
@@ -61,5 +61,11 @@ $utilities: (
     responsive: true,
     property: position,
     values: static relative absolute fixed sticky
+  ),
+  'max-width': (
+    responsive: true,
+    property: max-width,
+    class: mw,
+    values: (25: 25%, 50: 50%, 100: 100%)
   )
 );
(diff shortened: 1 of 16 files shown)


bikubi:
> [fcf2fdb](https://github.com/okfde/froide/commit/fcf2fdb855d94ac60a35b02eea83669d3b8168b7) is the meat here

...and `autoApprove` is not a great name...

frontend/styles/maps.scss
```diff
@@ -61,5 +61,11 @@ $utilities: (
     responsive: true,
     property: position,
     values: static relative absolute fixed sticky
+  ),
+  'max-width': (
```
krmax44, line 65: can't find these classes used anywhere. should we remove them then?

krmax44:
@pajowu @stefanw maybe take a quick look at the `fullscreen_app.html` template I introduced, which required some changes to the base template. https://github.com/okfde/froide/pull/876/files#diff-f523c28415b61abcd34ee758b144095d2b5819a2cd1070590d8d4978ffaf1200

frontend/styles/maps.scss
bikubi: they're here: https://github.com/okfde/froide/blob/bikubi/postupload-fixes/frontend/javascript/components/redaction/pdf-redaction.vue#L164 `mw-lg-50 mw-xl-25`

PR #883, opened 2024-10-30
fix postupload / plaintext get_differences default error

bikubi:


rugk:
Awesome! Thanks a lot for that super fast response!

PR #886, opened 2024-11-04
✨ allow specification of staff group instead of `User.is_staff` for `@is_staff`

krmax44:
`is_staff` is required for accessing the admin, which may also be non-staff


pajowu:
Should we use a different name than `is_staff` for this, since it has a clear meaning in the django-world?

krmax44:
Bit tricky, most synonyms are also used up (team, org). What about `is_crew`?

pajowu:
I like it

krmax44:
also moved it from auth helpers directly to the user model, take a look @pajowu

pajowu (approved):

PR #888, opened 2024-11-11
api for post upload

krmax44:
notable changes:

- split up `foirequest/api_views.py` into multiple files. this is a breaking change i.e. for froide-campaign and froide-food. checkout their respective branches for testing:
	- https://github.com/okfde/froide-food/tree/krmax44/foirequest-api
	- https://github.com/okfde/froide-campaign/tree/krmax44/foirequest-api
- add create/update methods to message endpoint
- modified `CreateOnlyWithScopePermission`, so that regular authentication can also be used. please review carefully
- added typescript api client, example:
	
	```typescript
	import { retrieveFoiMessage } from './api/index.ts'
	
	retrieveFoiMessage({
	  path: { id: '22699' }
	}).then((response) => {
	  console.log(response.data?.kind) // "email" | "post" | "fax" | "upload" | "phone" | "visit" | "import" | undefined
	})
	```
- added endpoints to message to convert a set of image attachments to pdf (previously part of a [non-api view](https://github.com/okfde/froide/blob/f827bcb7b0904c9c5f92d3098034f8bcc441c49a/froide/foirequest/views/message.py#L660))
- added delete endpoint for attachments


froide/foirequest/api/views/message.py
```diff
@@ -0,0 +1,55 @@
...
+        qs = get_read_foimessage_queryset(self.request).order_by()
+        return self.optimize_query(qs)
+
+    def optimize_query(self, qs):
+        return optimize_message_queryset(self.request, qs)
+
+    def ensure_draft(self):
+        # only drafts can be edited
+        if not self.get_object().is_draft:
+            raise PermissionDenied(_("Cannot alter non-draft messages"))
```
pajowu, line 47: This should probably be done using a permission-class. https://www.django-rest-framework.org/api-guide/permissions/#object-level-permissions

froide/foirequest/api/serializers.py
```diff
@@ -157,13 +161,16 @@ def create(self, validated_data):
...
+
 class FoiMessageSerializer(serializers.HyperlinkedModelSerializer):
-    resource_uri = serializers.HyperlinkedIdentityField(
-        view_name="api:message-detail", lookup_field="pk"
-    )
-    request = serializers.HyperlinkedRelatedField(
-        read_only=True, view_name="api:request-detail"
-    )
+    resource_uri = serializers.HyperlinkedIdentityField(view_name="api:message-detail")
+    request = FoiRequestRelatedField()
```
pajowu, line 173: is this intentional or should this be 
```suggestion
    request = FoiRequestRelatedField(read_only=True)
```

froide/foirequest/api/serializers.py
```diff
@@ -310,25 +348,13 @@ def get_file_url(self, obj):
...
+    )
     upload = serializers.CharField()
 
-    def validate_message(self, value):
-        writable_requests = get_write_queryset(
-            FoiRequest.objects.all(),
-            self.context["request"],
-            has_team=True,
-            scope="upload:message",
-        )
```
pajowu: what happened to this permission-checking / validation? Why is it not needed anymore?

froide/foirequest/api/views/attachment.py
```diff
@@ -43,8 +45,15 @@ class FoiAttachmentViewSet(
...
-    required_scopes = ["upload:message"]
+    required_scopes = ["make:message"]
+
+    def get_permissions(self):
+        if self.action == "list" or self.action == "retrieve":
+            return []
+        return [
+            CreateOnlyWithScopePermission(),
+            OwnsFoiRequestPermission("belongs_to__request"),
+        ]
```
pajowu, line 56: Shouldn't this action check happen in the permission-classes?

froide/foirequest/api/views/message.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def update(self, request: Request, *args, **kwargs):
+        self.ensure_draft()
+        return super().update(request, *args, **kwargs)
+
+    def destroy(self, request: Request, *args, **kwargs):
+        self.ensure_draft()
+        return super().destroy(request, *args, **kwargs)
```
pajowu, line 55: If `ensure_draft` is called via permission-class, these can be removed

froide/foirequest/models/message.py
```diff
@@ -37,10 +37,13 @@ def get_throttle_filter(self, queryset, user, extra_filters=None):
             qs = qs.filter(**extra_filters)
         return qs, "timestamp"
 
+    def with_drafts(self, drafts=True):
```
pajowu, line 40: Added here since I can't add it to the manager class directly:

Since we have a seperate proxy type for drafts, I would opt for not returning the drafts in the default manager (and the default related manager if easily possible)

froide/settings.py
```diff
@@ -501,7 +501,7 @@ def is_pkce_required(client_id):
             "read:email": _("Read user email"),
             "read:request": _("Read your (private) requests"),
             "make:request": _("Make requests on your behalf"),
-            "upload:message": _("Upload postal messages"),
+            "create:message": _("Add messages and attachments to user's request"),
```
pajowu, line 504: Could this lead to weird behaviours for existing keys / apps that don't have this scope?

froide/foirequest/models/message.py
```diff
@@ -64,12 +67,21 @@ class MessageKind(models.TextChoices):
...
     PHONE = ("phone", _("phone call"))
     VISIT = ("visit", _("visit in person"))
     IMPORT = ("import", _("automatically imported"))
 
 
+# users are allowed to only create messages of these kinds
+# the other kinds can only be created by the system
+MESSAGE_KIND_USER_ALLOWED = [
+    MessageKind.POST,
+    MessageKind.PHONE,
```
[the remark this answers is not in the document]
krmax44: There already exist multiple messages of kind `PHONE` (mostly staff requests or kind manually changed by staff).

Maybe we could actually add the possibility for users to add e.g. call notes. In Berlin for example, fees are often waived when the information can be provided via a phone call.

froide/foirequest/api/permissions.py
```diff
@@ -0,0 +1,16 @@
+from rest_framework import permissions
```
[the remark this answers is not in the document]
krmax44: So better to put the api views in an `api_views/` module and everything else in the app root?

froide/foirequest/api/permissions.py
```diff
@@ -0,0 +1,16 @@
...
+class OwnsFoiRequestPermission(permissions.BasePermission):
+    def __init__(self, request_field: str = "request") -> None:
+        self.request_field = request_field
+        super().__init__()
+
+    def has_object_permission(self, request: Request, view, obj):
+        if request.method in permissions.SAFE_METHODS:
+            return True
+
+        return can_write_foirequest(getattr(obj, self.request_field), request)
```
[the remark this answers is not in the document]
krmax44: Seems to be the case. Somehow worked in my initial testing, but I should probably split this class up for Message and Attachment anyhow, since the check for the `create` method also differs quite a bit.

froide/settings.py
krmax44: Yes. I suppose it's only used by the scanner app though.

froide/foirequest/models/message.py
krmax44: That would require a seperate API endpoint for message drafts as well then. Since we're also splitting up the admin for drafts and non-drafts, I suppose it's okay to do that? It just feels a little repetitive to me, since they only differ in one attribute.

froide/foirequest/api/serializers.py
krmax44: Yes, since `read_only` also affects the `create` action. I've added a safeguard to prevent the field from being modified: [`eab4414` (#888)](https://github.com/okfde/froide/pull/888/commits/eab44143d4fba54048cee05f758d748e1bf0852c#diff-942b7081675f4bd4e08c833a5d7cc2cf9d6631a444a5d0419f5616c7a5dbf7e4R178-R181) (see also https://github.com/encode/django-rest-framework/discussions/8606)

froide/foirequest/api/permissions.py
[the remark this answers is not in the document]
krmax44: Updated `WriteFoiRequestPermission` like so:

https://github.com/okfde/froide/blob/b5e31e9dce92f6aca4d99200d0acc87d30e90d75/froide/foirequest/api/permissions.py#L11-L19

froide/foirequest/api/permissions.py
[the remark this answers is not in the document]
pajowu: i think this is the usual structure, but you can keep it as is too

froide/settings.py
krmax44: Only staff users have tokens with this scope.

froide/foirequest/api/serializers.py
pajowu: I think it might be better to work with an allowlist here (aka only allow changing this on creation). Does `update` also cover `partial_update`)?

froide/foirequest/models/message.py
pajowu: As discussed today, let's add two endpoints

krmax44:
I've split up models and views by `FoiRequestMessage` and `FoiRequestDraftMessage` (with endpoints `api/v1/message` and `api/v1/message/draft`). While this also reduces the risk of leaking drafts, in the request's detail page the `message` attribute will now no longer list drafts. If that is undesirable, we'd need to update this function: https://github.com/okfde/froide/blob/f00b8e1c2c53f240393cc260f3b37eed1a50748c/froide/foirequest/serializers.py#L138-L144

froide/foirequest/models/message.py
```diff
@@ -37,10 +37,13 @@ def get_throttle_filter(self, queryset, user, extra_filters=None):
...
 
+    def with_drafts(self, drafts=True):
+        return super().get_queryset().filter(is_draft=drafts)
 
-class FoiMessageNoDraftsManager(FoiMessageManager):
+
+class FoiMessageDraftManager(FoiMessageManager):
     def get_queryset(self):
-        return super().get_queryset().filter(is_draft=False)
+        return self.with_drafts()
```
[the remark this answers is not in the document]
krmax44: This should be fixed now.

froide/foirequest/api/serializers.py
```diff
@@ -0,0 +1,382 @@
...
+                _("You do not have permission to add a message to this request.")
+            )
+        return value
+
+    def ensure_draft(self, instance):
+        if not instance.is_draft:
+            raise serializers.ValidationError("Only draft messages can be altered.")
+
+    def create(self, validated_data):
+        return super().create(validated_data)
```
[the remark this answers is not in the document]
krmax44: A signal is now sent when a draft message is published.

frontend/javascript/api/index.ts
```diff
@@ -0,0 +1,12 @@
+import { client } from './gen/services.gen'
+
+const token = document.cookie.match(/csrftoken=([^;]+)/)?.[1]
+
+client.interceptors.request.use((request) => {
+  request.headers.set('X-CSRFToken', token!)
```
krmax44, line 6: don't set this header when method is get

pajowu (approved):

stefanw (approved):

PR #889, opened 2024-11-11
mobile friendly pdf redaction

bikubi:
At the moment this contains pdfupload-fixes; I will rebase as soon as that lands.

- pinch-zoom libraries evaluation
    - <https://github.com/coddicat/vue-pinch-scroll-zoom>
        - made a MWE, half worked
        - requires content and containr sizes provided in absolute pixels
          responsive sizes only via [workaround](https://gist.github.com/qq7/d0ded29cf75572dce06a2c3e7478d92f)
    - <https://github.com/anvaka/panzoom>
        - half implemented, worked
        - seems alive, but old (not a real module...)
        - rather big
        - [no pan-while-pinch-zooming](https://github.com/anvaka/panzoom/issues/307)
    - <https://github.com/drozhzhin-n-e/vue-pinch-zoom>
        - didnt try, last commit 4y ago
    - <https://github.com/GoogleChromeLabs/pinch-zoom>
        - felt awkward, didnt pursue further
    - <https://github.com/manuelstofer/pinchzoom>
        - many stars
        - lightweight, smol
        - but is [looking for a maintainer](https://github.com/manuelstofer/pinchzoom/issues/126)
        - old, [doesnt use PointerEvents](https://github.com/manuelstofer/pinchzoom/issues/114)
    - settled on <https://github.com/timmywil/panzoom>
        - lightweight, tiny
        - uses modern PointerEvents
        - supports pan-while-pinch-zooming
        - works ootb with mouse
- optionally considered: on mobile, pan two-fingers only, one finger = auto paint + get rid of tool toggle
  ⇒ abandoned, because code was behaving weirdly
- I've added maps style zoom buttons
- added "hi dpi render scale"
    - PDF previews were bitmap rendered at device resulution, became very pixelated at even low zoom 
    - currently set a 2x, but could depend on display size (but probably something smarted, capped at, like 2k)

note the tool toggle on the left and the zoom ui:

grafik

~~last minute TODO: the tooltip I've added to the tool toggles doesn't respect `data-bs-placement` (I don't know why), so I might just disable the tooltip on mobile, where it can block the other button, and is generally in the way.~~

- `placement` does work, it just flips to `right` for the long `title`s
- I've implemented a way to auto-hide the tooltips after a timeout (set to 3s), after `focusin` events - this won't affect mouse users, but mobile touchers and (unfortunately) keyboard users. A cleaner fix would send adequate events from panzoom to `focusout` the tooltips...


bikubi:
@pajowu @krmax44 
toolbars are now sticky, i moved the two main tools there, moved pagination to the bottom toolbar, which relaxed everything. i've removed the tooltips, they didn't make sense (as a replacement for button labels on mobile), instead i've updated the how-to above, directly&visually referring to the buttons by label or icon, resp.

krmax44:
Also, the help texts for redaction differ between the "standalone" redaction and when you're redacting during post upload. We should probably use the same on both. Maybe with the option to collapse them, so experienced users don't have to scroll as much.

krmax44:
@bikubi I added switching to the move tool by holding down the Alt key and dragging. What do you think?

krmax44 (approved):
This is such a huge improvement, especially on mobile. Only two things left from my point of view:

- Remove container (at least on mobile, the horizontal padding looks odd and wastes precious space)
- Possibly add end-to-end tests (would be nice since this is a pretty critical feature, but def. not merge blocking imho)

bikubi:
@krmax44 

> I added switching to the move tool by holding down the Alt key and dragging. What do you think?

1. As I've said before, not a fan, *personally* :) It interferes with the somewhat standard (but maybe old school) move-window modifier on various Linux-adjacent DEs, plus users of Win+Mac apps which provide this functionality. But this might affect very few users.
2. So to test, I've disabled Alt-move on my machine and there's another thing. I believe most browsers come with the menu bar (File, Edit etc.) hidden by default (on Win and Linux). A tap on Alt focuses and unhides the menu bar, shifting the whole viewport down (yes, this collides with behavior (1.) and, yes it is annoying)
3. Unlikely, but there might be long-press behavior that we are not aware of (like Photoshop's Alt+Shift-drag changing Windows's keyboard layout...)

It's tough! There were similar issues with Shift, Ctrl and combinations thereof.

If you want to go ahead and we find a solution, let's remember to mention it in the on-line help (I had it in a `bs-tooltip` on the move button; later moved to the `<detail>` instructions.) 

> This is such a huge improvement, especially on mobile. Only two things left from my point of view:
> 
>     * Remove container (at least on mobile, the horizontal padding looks odd and wastes precious space)

This gets awkward on big screens (which tend to be wide). I assume the most common document is portrait A4. Currently, the "canvas viewport" auto-sizes and prefers "tallness". On a 16:9 screen, this means that ~¾ of the document will initially appear below the fold. Not terrible, but awkward IMO.
Also awkwardly, the "zoom fit" button does not work like a "zoom to page" button, since again, we both dimensions to the (unexpectedly) tall container.
I'm not sure there's a perfect solution here. We could limit the "canvas viewport container" to `100vh` and prefer panning over scrolling, but this might only shift the awkwardness to other places.

For now, I've made the container `-xxl` which makes it nicely wide on mobile but avoids super-wide awkwardness.
If you want to give it a try yourself, just remove the `container-xxl` classes and, you know, maximize on a big monitor and close the right pane DevTools :)

>     * Possibly add end-to-end tests (would be nice since this is a pretty critical feature, but def. not merge blocking imho)

I would like to get this out of the door rather sooner than later for the new attachment "manager", but let's discuss later?
There are also issues with the text mode...

While testing the wider container I've fixed more bugs -- see my last commits.

PR #906, opened 2024-11-19
🚸 Do not strip leading / trailing spaces in foirequest body

pajowu:
This breaks leading indentation, e.g.

Please send me the following:
"  1. Document A
  2. Document B"


stefanw (approved):
I'm not sure if we want that. Maybe still strip leading/trailing new lines? Not feeling strongly about this though.

PR #912, opened 2024-11-25
🍱 Remove twitter

pajowu:


krmax44 (approved):

krmax44:
FYI: There were only two relevant pages using the plugin, which don't conflict with this change.

PR #915, opened 2024-11-25
♻️ Replace pygtail with own alternative

pajowu:
This PR replaces the pygtail dependency with a smaller version I wrote based on pygtail. This version notably does not include a racecondition that pygtail has, which breaks our logparsing on every logrotate until manual intervention


stefanw (approved):
Let’s merge it before Sunday 😉

Makefile
```diff
@@ -18,5 +18,5 @@ messagesde:
 	python manage.py makemessages -l de --ignore public --ignore froide-env --ignore node_modules --ignore htmlcov --add-location file
 
 requirements: pyproject.toml
-	uv pip compile -o requirements.txt pyproject.toml
-	uv pip compile -o requirements-test.txt --extra test pyproject.toml
+	uv pip compile -o requirements.txt pyproject.toml -p 3.10
```
krmax44, line 21: Is there a way to put that into `pyproject.toml`?

compose-dev.yaml
```diff
@@ -8,7 +8,7 @@ services:
       POSTGRES_DB: froide
       POSTGRES_PASSWORD: froide
     ports:
-      - 127.0.0.1:5432:5432
+      - 127.0.0.1:5433:5432
```
krmax44, line 11: Is this change intentional?

pajowu: oops, no

Makefile
pajowu: There is, but it brakes the CI

PR #941, opened 2025-01-22
🚸 Always allow mods to change request status

pajowu:
Takes the easy part from https://github.com/okfde/froide/pull/865 to get it merged sooner


stefanw:
Sorry, I (or we?) accidentally merged this by setting the branch settings temporarily too lax and it was already on 'merge when all-checks-pass.

Do we really want mods to change status of any request visible to them immediately? Does that prevent users from learning how to do that themselves? What if they don't agree with the status and wonder where it came from? Previously it was only possible after 31 days which seemed reasonable.

PR #947, opened 2025-02-12
Add improvement to OAuth flow

stefanw:


froide/account/views.py
```diff
@@ -154,6 +154,9 @@ def confirm(
 
 
 def go(request: HttpRequest, user_id: str, token: str, url: str) -> HttpResponse:
+    if url == "/":
```
pajowu, line 157: `== "/"` or `== reverse("index")`?

stefanw: It should be "/" because that denotes an 'empty' URL match at the end of this regex path:

https://github.com/okfde/froide/blob/main/froide/account/urls.py#L63

(A once elegant URL path design that turned peculiar with time 😛)

pajowu: 🤯

PR #949, opened 2025-02-13
Add postupload app integration

stefanw:
This adds:

- a mobile app reference loaded from a URL with the message id (configurable via settings)
- a Websocket connection that listens for new attachments added to the message (e.g. via the app)


stefanw:
It got a bit messy when the new FoiAttachment post_save signal triggers an async event loop which broke the tests due to other event loops. I got the test suite fixed, the other parts should be relatively straight forward.

PR #959, opened 2025-02-24
more foirequest api improvements

krmax44:
- non-draft, postal messages can now also be updated via API (some fields only, like `registered_mail_date`)   
  - maybe todo for later: allow changing subject + content
- approve/unapprove attachments via api
- attachment to document via api
- allow updating foirequest (resolution, law...) via api


stefanw:
Adding an attachment via upload to a *draft* message through the API is currently not working:
See test case here:

https://github.com/okfde/froide/commit/f1f97ad68168d0cb6104f8aac4aaab46445c3600

Looks like the `FoiMessageRelatedField` is trying to validate the `view_name` early and never gets to the `get_url` logic that branches on object type.

froide/foirequest/api_views/message.py
```diff
@@ -75,18 +78,6 @@ def publish(self, request, pk=None):
         if not message.is_response:
             message.sender_user = request.user
```
stefanw, line 79: Considering we have teams with write access this should not be set to the acting user. Maybe this field is actually not needed as in a request the only non-public body sender can ever officially be the requester? To record who actually acted on the request, we should use `FoiEvent` logging.

stefanw (changes requested):
Does this have enough `FoiEvent` logging? Do we need new event types, e.g. for attachment approval? Given these new APIs being able to track how a change happened will be important.

krmax44:
> Does this have enough `FoiEvent` logging? Do we need new event types, e.g. for attachment approval? Given these new APIs being able to track how a change happened will be important.

I've added `FoiEvent` logging in a few places where I've missed them. Also, I created a new `ATTACHMENT_CONVERTED` event type (for image/docx/... → pdf). I wanted to save the originals in the event context (`{"originals": [425451, 425452]}`), but the data is stringified (yielding `{"originals": "[425451, 425452]"}`). Is this required?

https://github.com/okfde/froide/blob/a4d4d6b2d792873dc23a7d67bf7867dfbe762a77/froide/foirequest/models/event.py#L148

For draft messages, I think we can omit events until they are published and become a "real" message. Do you see downsides with that?

krmax44:
Should we log if changes via the api were made using non-official oauth applications?

stefanw:
> I wanted to save the originals in the event context ({"originals": [425451, 425452]}), but the data is stringified (yielding {"originals": "[425451, 425452]"}). Is this required?

I think there were possibly occasions when the context contained types that can't be JSON encoded automatically (e.g. `datetime`s). `str` ensures the saving works.

PR #960, opened 2025-02-24
New Attachment Manager

bikubi:


frontend/javascript/components/docupload/attachment-actions.vue
```diff
@@ -0,0 +1,278 @@
...
+    @click="redactClick($event, unredacted)"
+    >
+    <i class="fa fa-pencil-square"></i>
+    {{ i18n.editRedaction }}
+  </a>
+  <button v-if="attachment.is_irrelevant" @click="makeRelevantClick" type="button" class="btn btn-sm btn-link text-start">
+    <i class="fa fa-exclamation-circle"></i>
+    {{ i18n.markNotIrrelevant }}
+  </button>
+  <a class="btn btn-sm btn-link text-start" :href="attachment.file_url" download>
```
krmax44, line 137: Right now, the default file names when downloading are the file hashes. Better to provide the name:

```suggestion
  <a class="btn btn-sm btn-link text-start" :href="attachment.file_url" :download="attachment.name">
```

(more of this below)

krmax44:
When uploading an image, and then immediately converting the image to a pdf, the loading spinner (maybe the entire page?) freezes.

krmax44:
Bulk deleting attachments should require a confirmation

frontend/javascript/components/bs-modal.vue
```diff
@@ -0,0 +1,89 @@
...
+  <Teleport to="body">
+    <div
+      ref="modalEl"
+      class="modal"
+      tabindex="-1"
+      role="dialog"
+      v-if="doRender"
+      >
+      <div
+        :class="'modal-dialog ' + dialogClasses"
```
krmax44, line 68: Instead, do this:


```suggestion
        v-bind="$attrs"
```

This will bind all (non-prop) attributes to the element. So not just classes work!

frontend/javascript/components/bs-toast.vue
```diff
@@ -0,0 +1,35 @@
...
+  teleportTo = document.createElement('div')
+  teleportTo.id = 'toastContainer'
+  teleportTo.classList = 'toast-container position-fixed bottom-0 end-0 p-3'
+  document.body.appendChild(teleportTo)
+}
+</script>
+
+<template>
+  <Teleport :to="teleportTo">
+    <div v-bs-toast :class="`toast show text-bg-${props.color}`" role="alert" aria-live="assertive" aria-atomic="true">
```
krmax44, line 26: Low prio/for later: The toast currently looks a bit odd to me. Shadow seems to far from the element. Also, it's tiny and easy to miss on a larger screen. Maybe it needs a stronger animation instead of the fade.

frontend/javascript/components/docupload/attachment-manager.vue
```diff
@@ -0,0 +1,117 @@
...
+provide('config', props.config)
+
+const { attachments, addFromUppy, refresh } = useAttachments({
+  urls: {
+    ...props.config.url,
+    getAttachment: props.config.url.getAttachment.replace('/0/', '/') +
+      '?belongs_to=' +
+      props.message.id,
+  },
+  csrfToken: document.querySelector('[name=csrfmiddlewaretoken]').value,
```
krmax44, line 29: Maybe the api module should export csrfToken, so it can be used directly here (single source of truth).

frontend/javascript/components/docupload/attachment-manager.vue
```diff
@@ -0,0 +1,117 @@
...
+
+watch(
+  () => attachments.images.length,
+  (newValue) => {
+    if (newValue > 0) {
+      nextTick().then(() => {
+        imagesConverterContainer.value?.scrollIntoView({ behavior: 'smooth', block: 'nearest' })
+      })
+    }
+  }
```
krmax44, line 45: Bit of a preference thing, but I'd prefer:


```suggestion
  async (newValue) => {
    if (newValue > 0) {
      await nextTick()
      imagesConverterContainer.value?.scrollIntoView({ behavior: 'smooth', block: 'nearest' })
    }
  }
```

Less nesting usually means more readable imho

frontend/javascript/components/docupload/attachments-table.vue
```diff
@@ -0,0 +1,304 @@
...
+      <button type="button" class="btn btn-link" @click="selectAll">
+        {{ i18n.selectAll }}
+      </button>
+      <button type="button" class="btn btn-link" @click="selectNone">
+        {{ i18n.selectNone }}
+      </button>
+    </div>
+  </div>
+
+  <!-- AS CARDS -->
```
krmax44, line 156: low prio/for later: maybe let the user also choose between the two views

frontend/javascript/components/docupload/image-page.vue
```diff
@@ -1,35 +1,146 @@
+<script setup>
```
krmax44, line 1: Low prio/later: We should probably improve this a bit. The "cut" button feels very unintuitive and it can't be undone without refreshing. Also the animation when rotating 360 degrees is very unsatisfying 😉

frontend/javascript/components/docupload/attachments-table.vue
```diff
@@ -0,0 +1,304 @@
...
+        type="checkbox"
+        ref="selectAllEl"
+        :checked="selectAllState"
+        :indeterminate.prop="selectAllState === undefined"
+        @click="selectAllClick"
+        />
+      <span class="sr-only">{{ selectAllLabel }}</span>
+    </label>
+    <div v-if="selectionActions" class="d-flex flex-grow-1 justify-content-end">
+      <!-- TODO, needs backend support
```
krmax44, line 137: Maybe also bulk "mark as result"?

frontend/javascript/components/docupload/attachment-actions.vue
bikubi: when I test this locally, both FF and Chromium ignore this, and prefer the filename from the endpoint's header:

`http://localhost:8000/files/foi/900123/brief_geschwaerzt.pdf?token=f0ob4r` ⇒  
`Content-Disposition: inline; filename="d3adb33f….pdf"`

krmax44: Oh, didn't realize we set that using headers.

frontend/javascript/components/bs-modal.vue
bikubi: I'd add the `v-bind` if it were the outermost `<div>`, the one with `class="modal"`, to allow `<bs-modal class="fade"`... there is also `contentClasses` right after it. yagni? too much bike shedding?

frontend/javascript/components/bs-toast.vue
bikubi: yes, the default color one is very faint. errors have bg-danger.. maybe make the default bg-primary. in the BS examples, the primary avatary icon also helps...

frontend/javascript/components/online-help.vue
```diff
@@ -0,0 +1,81 @@
...
+        v-show="isFetching">
+        <span class="sr-only">{{ i18n.attachmentsLoading }}</span>
+      </span>
+      <div v-html="contents"></div>
+    </template>
+  </bs-modal>
+</template>
+
+<style lang="scss">
+// unscoped, because getting scoped to work thorugh BsModal's Teleport is hard or not yet possible
```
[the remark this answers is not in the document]
bikubi: TIL about BS Offcanvas, I had been wondering what they are doing with [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas)...

frontend/javascript/components/docupload/attachments-table.vue
bikubi: fixed by [f70d7d8](https://github.com/okfde/froide/pull/960/commits/f70d7d84ed3488cb52e2c0917a6ac32958bf1c77)

frontend/javascript/components/docupload/attachment-manager.vue
bikubi: [3f7bc63](https://github.com/okfde/froide/pull/960/commits/3f7bc632ec42b57655ae7b5f9a9a3df6932487b7)

frontend/javascript/components/docupload/attachment-manager.vue
bikubi: fixed in 8eabdd0cf8eb79bdc3d9f8fac6c1552302010f5e

frontend/javascript/components/docupload/attachment-actions.vue
bikubi: attempted in 54f574ad3169c401f43329d63a24d005f925f83b if you want to give it a try

bikubi:
@krmax44 yikes! found a lot of bugs while going through your review, and quick changes turned out substantial, but worthwhile, I believe.

I'd appreciate (quick) review of the many little fixes & improvements since, but most notably d5ca2221e26cdb4cc197f90a8e365d85e83a0538, where I'd *especially* appreciate another look on the Django parts & the awkward "`$root` providing wrapper" (yet another case of Vue2/3 interop; the alternative would be to Vue3ify django-filingcabinet...)

While putting `<document-viewer>` into `<attachment-icon-preview>` I ran into a weird caching bug, that *might* be a browser bug:

1. in new Attachment Management, redact & publish ("Ergebnis markieren") an Attachment. It becomes a Document.
2. Re-redact the original Attachment PDF, by clicking its "Redact/Schwärzen" action button. A modal with `<pdf-redactions>` should open. Apply the redactions; modal should close.
3. Click the Document's preview icon to "inspect the redaction" you've just done
4. Observe that the preview is out of date.
    - It refreshes after a page reload (or context-click-open-in-new-tab on the image). So likely this wasn't a problem when this wasn't "modals in an SPA-ish" yet.
    - Source are the `<picture><source …>`s [in django-filingcabinet/…/document-page.vue](https://github.com/okfde/django-filingcabinet/blob/main/frontend/javascript/components/document-page.vue#L6-L41); URLs look like `/files/docs/ab/cd/ef/abcdef1234…/page-p1-small.png.webp`, judging by Chromium's log, but...
    - Observed in Firefox, not in Chromium. Firefox also does not show the images in DevTools Network tab.

(On top of all that I'm not 100% sure if Documents are even supposed to be re-redactable?)

krmax44:
@bikubi Yes, documents should be re-redactable (that is, the attachment is re-redacted, and the document "recreated" with the resulting PDF).

I can't reproduce your findings, since I'm experience a different set of bugs:

- When redacting a document (attachment "als Ergebnis"), using the "Edit redaction" button in the overview (so not within the preview modal), ...
	- ...it works fine the first time
	- But when redacting it again right after, trying to open the preview doesn't work (in a very weird way: sometimes opening a preview of another attachment works, sometimes it does not)
- When redacting a document within the preview modal, I get stuck with a modal backdrop overlaying the screen (but with no modal visible)

Some things to keep in mind:

- Reprocessing the document may take some time. Whenever `document.pending` is `True`, the preview should probably display the PDF as if it were a normal attachment (see https://github.com/okfde/django-filingcabinet/blob/554cf48c92fbf1f597301afedb84275c75fde8c8/src/filingcabinet/templates/filingcabinet/_document_viewer.html#L147-L153)
	- so after redacting a document, maybe "lock" the attachment from editing until it's processed? Doesn't need to update automatically/check if the document is finished, page reload for that is probably fine for now.
- since the URLs of the PDF's page images contain a hash, they change when the content changes. Caching issues should be unlikely.

PR #992, opened 2025-04-15
Check that user entered pk in make_choose_object_action

noeddl:
See https://github.com/fragdenstaat/issues/issues/18

### DIFF
--- a/froide/helper/admin_utils.py
+++ b/froide/helper/admin_utils.py
@@ -44,6 +44,13 @@ def make_choose_object_action(model_or_queryset, callback, label):
                 callback(model_admin, request, queryset, action_obj)
                 model_admin.message_user(request, _("Successfully executed."))
                 return None
+            else:
+                model_name = model._meta.verbose_name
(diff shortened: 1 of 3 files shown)


noeddl:
```python
def make_choose_object_action(model_or_queryset, callback, label):
    if issubclass(model_or_queryset, models.Model):
        model = model_or_queryset
        filter_qs = None
    else:
        filter_qs = model_or_queryset
        model = model_or_queryset.model
    ...
```

I didn't find any calls to `make_choose_object_action` where `model_or_queryset` in fact is a QuerySet. So maybe this could be simplified to only accept a Model subclass? Also, I think this check does not work as expected which I realized while writing tests (see https://github.com/okfde/froide/pull/992/commits/8edac82d738160c689ac0cef18c7552890db2cab). I suppose the idea is to either use a Model subclass or a QuerySet instance. But the way it is written now you would have to use a QuerySet subclass for it to work (otherwise `issubclass()` fails) which probably is not what is intended?

krmax44 (approved):

PR #1003, opened 2025-04-29
Resolve datetime deprecation warnings

emmanuel-ferdman:
# PR Summary
This small PR resolves the deprecation warnings on `datetime` in Python3.12+. You can find them in the [CI logs](https://github.com/okfde/froide/actions/runs/14664674665/job/41156572872#step:12:134):
```python
/home/runner/work/froide/froide/froide/account/tests/test_account.py:131: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

/home/runner/work/froide/froide/froide/helper/spam.py:196: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

### DIFF
--- a/froide/helper/spam.py
+++ b/froide/helper/spam.py
@@ -2,7 +2,7 @@ import logging
 import os
 import re
 from dataclasses import asdict, dataclass
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Optional, Set
 
 from django import forms
@@ -193,7 +193,7 @@ class SpamProtectionMixin:
             )
         if self._should_include_timing():
             self.fields["time"] = forms.FloatField(
-                initial=datetime.utcnow().timestamp(), widget=forms.HiddenInput
+                initial=datetime.now(timezone.utc).timestamp(), widget=forms.HiddenInput
             )
 
     def _should_include_timing(self) -> bool:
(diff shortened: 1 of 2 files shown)


froide/helper/spam.py
```diff
@@ -232,7 +232,7 @@ def clean_test(self) -> str:
 
     def clean_time(self) -> str:
         value = self.cleaned_data["time"]
-        since = datetime.utcnow() - datetime.fromtimestamp(value)
+        since = datetime.now(timezone.utc).replace(tzinfo=None) - datetime.fromtimestamp(value)
```
pajowu, line 235: why the replace?

emmanuel-ferdman: @pajowu  The `replace(tzinfo=None)` strips the UTC timezone info from `datetime.now(timezone.utc)`, making it a naive datetime (like the original `datetime.utcnow()`). Without this, subtraction fails with `TypeError` because `fromtimestamp()` is timezone-naive while `datetime.now(timezone.utc)` isn't. We can also pass `timezone.utc` to `fromtimestamp()` and remove the `replace(tzinfo=None)` but `replace(tzinfo=None)` keeps  the original naive behavior of `since`.

stefanw: I think it makes more sense to compare two aware datetimes instead.

stefanw:
Thanks for your work. Superseded by https://github.com/okfde/froide/pull/1022

PR #1022, opened 2025-05-21
Upgrade to Django 5.2

stefanw:
- Resolves some deprecation warnings through upgrades or removal
- Removes no longer maintained `django-fsm` and replaces it with basic state transitions in Upload model
- Keeps legacy password hash wrapper working. This should be removed soon and remaining passwords marked unusable.


stefanw:
Currently blocked on https://code.djangoproject.com/ticket/34898 missing a workaround.

stefanw:
Some hacking on old migrations was necessary due to a different behaviour when migrating fields with non-deterministic collations (here: case insensitive email field).

PR #1061, opened 2025-07-07
fix date validation for post uploads on the same day the request was created

bikubi:
example:
- create foirequest on 2025-07-07 13:00
- in postupload flow, set date 2025-07-07
- `timestamp` becomes 2025-07-07 12:00 (default is noon)
- this throws a `ValidationError Your message date is before the request was made.` [here](https://github.com/okfde/froide/blob/19a5948b9f34b6c84a8f67fd3d7bd4922c4d8c25/froide/foirequest/serializers.py#L402)

in this case, we do not set noon, but the request's created_at


frontend/javascript/components/postupload/post-upload.vue
```diff
@@ -171,8 +171,15 @@ const isoPrepareDate = (dateYmdStr) => {
...
   }
-  // Note that this could lead to unexpected behavior between
+  // Note that the logic above could lead to unexpected behavior between
   // "client midnight" and "server midnight".
+  // Last, we avoid inventing a too early timestamp,
+  // (request created in the pm, post uploaded same day)
+  const requestCreatedAt = new Date(props.foirequest.created_at)
+  if (dateTime < requestCreatedAt) {
+    // "round up" a second, avoid precision collision
+    dateTime = new Date((+requestCreatedAt) + 1000)
```
bikubi, line 181: @krmax44 is this too cursed? :)

PR #1069, opened 2025-07-14
postupload: retry failed images pollingly, avoid websocket race condition

bikubi:


bikubi:
fyi @krmax44 not pretty, but i'd like to check if it works at all

PR #1106, opened 2025-08-26
🚸 Show deferred message body in admin

pajowu:


froide/foirequest/admin.py
```diff
@@ -1021,8 +1021,57 @@ class DeferredMessageAdmin(admin.ModelAdmin):
...
 
-    save_on_top = True
+    def decoded_mail(self, obj):
+        return format_html(
+            "<pre>{}</pre>",
+            obj.decoded_mail(),
+        )
+
+    def failed_authenticity(self, obj):
+        with closing(BytesIO(base64.b64decode(obj.mail))) as stream:
```
stefanw, line 1064: Nit, but maybe we move email parsing logic into the model and cache the parsed email on the model object? Then we don't need to parse multiple times.

stefanw (approved):

PR #1113, opened 2025-09-02
♻️ Ensure usernames and secret addresses are lowercase

pajowu:


stefanw:
So for email addresses that we control, I would generally advise against changing the db collation. We did it for the user's account email address because – technically – `Test@example.org` and `test@example.org` are different e-mail-addresses and we want to store and use them as given by the user and at the same time:
 - disallow the same address with different casing for a different user.
 - matching the user when entered with a different case.

However, the email field can then no longer be easily used in Django's ORM (e.g. admin search fields) because it needs [an annotation to be filtered](https://github.com/okfde/froide/blob/7fe1d35f87bc2a97a70d907078f45c7bee1ef350/froide/account/models.py#L50-L51). This is an added complexity that has its use but that I'd rather keep contained to where it makes sense.

These email fields we generally can ensure lower casing in the DB and when we do queries against them, so my instinct is that the different collation is too much.

I'm wondering if instead we should just make the fields in question read-only in the admin (good idea regardless) and/or add model validations to ensure lower casing. It's not a panacea, but the current system has worked for a long time and it just seems that the current misbehavings are the result of a staff admin input?

PR #1127, opened 2025-10-01
fixes for PDF redaction: browser page zoom, rotated PDFs, text-mode

bikubi:
tricky, tricky stuff

### DIFF
--- a/frontend/javascript/components/redaction/pdf-redaction.vue
+++ b/frontend/javascript/components/redaction/pdf-redaction.vue
@@ -655,10 +655,11 @@ export default {
       return doc.getPage(pageNum).then((page) => {
         this.page = page
 
-        let maxWidth = this.$refs.containerWrapper.offsetWidth
         if (!this.$refs.containerWrapper) {
           console.error('containerWrapper is null?')
+          return
         }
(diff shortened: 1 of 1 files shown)


frontend/javascript/components/redaction/pdf-redaction.vue
```diff
@@ -648,8 +667,13 @@ export default {
         this.page = page
 
         let maxWidth = this.$refs.containerWrapper.offsetWidth
+        if (!this.$refs.containerWrapper) {
```
krmax44, line 670: This code will never be reached, as the line above would cause a `TypeError` should `this.$refs.containerWrapper` be null-ish.

frontend/javascript/components/redaction/pdf-redaction.vue
```diff
@@ -149,6 +150,7 @@
 
         <div class="btn-group me-1 toolbar-modes py-2">
           <button
+            type="button"
```
krmax44, line 153: Let's maybe switch to a [checkbox toggle button](https://getbootstrap.com/docs/5.3/forms/checks-radios/#checkbox-toggle-buttons) here as well

PR #1179, opened 2026-01-22
🔧 Ensure that debug_toolbar is in INSTALLED_APPS before adding its URLs

noeddl:
Otherwise, the dev server will crash.


krmax44:
Maybe we can simply add the app (and the middleware) to the `Dev` settings class. Might require some changes downstream in the fds_de settings though.

noeddl:
@krmax44 Good idea, I'll take a look at solving it this way.

PR #1243, opened 2026-04-13
Use pytest-xdist with one group and two workers

stefanw:
Tests that need elasticsearch or playwright are grouped into one worker to run sequentially.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
         run: pnpm run build
       - name: Run tests
         run: |
-          coverage run --branch -m pytest froide/
+          pytest -n auto --cov --cov-report= froide/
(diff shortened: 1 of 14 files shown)


krmax44:
Great! I tried this approach before too, but seemed to have missed some tests for the sequential group. Seems to work reliably now, also with more than two workers.

Minor things:
- I removed the `-n` option from `pyproject.toml` and set it explicitly in CI (to `-n auto`), as always running parallelized  can cause issues with tools like the [pytest debuger within VS Code](https://github.com/ernestoarbitrio/pytest-runner).
- I also removed the warning filters. We're currently getting only four unique warnings, at least one of which is useful.
- I added pytest-cov, which properly reports coverage with parallelized runs (before, CI reported only 16 %).

PR #1247, opened 2026-04-15
Fix spurious language field migrations and add USER_LANGUAGES

noeddl:
- Replace inline use of `settings.LANGUAGE_CODE` and `settings.LANGUAGES` with callables so changes to these settings no longer trigger new migrations (see also https://github.com/okfde/django-filingcabinet/pull/101).
- Add new settings variable `USER_LANGUAGES` to allow for defining a separate set of languages to be used in user dialogues, e.g. to exclude `de-ls` from user language choice in the account settings.


noeddl:
Tests fail because https://github.com/okfde/django-filingcabinet/pull/101 has to be merged first.

froide/account/forms.py
```diff
@@ -625,7 +626,11 @@ def save(self, commit=True):
 
 
 class AccountSettingsForm(forms.ModelForm):
+    language = forms.ChoiceField(
+        choices=get_user_language_choices,
```
n0ctua, line 630: What happens here if a stored value is no longer selectable? If somehow a user picked a language from LANGUAGES, that is later excluded via the new USER_LANGUAGES than the account settings form becomes invalid. 

Do we need a guard function that sets the selected language back to a default when USER_LANGUAGES changes in the future or would that be addressed by a migration that changes the exisiting values in the database?

froide/foirequest/forms/request.py
```diff
@@ -114,7 +115,7 @@ class RequestForm(JSONMixin, forms.Form):
     )
     tags = TagField(required=False, widget=forms.HiddenInput)
     language = forms.ChoiceField(
-        choices=settings.LANGUAGES,
+        choices=get_user_language_choices,
```
n0ctua: Is it intended to limit this hidden field to USER_LANGUAGES or is it more of an UI concept for pickers etc so this should still be settings.LANGUAGES?

froide/helper/language.py
```diff
@@ -0,0 +1,13 @@
...
+def get_default_language():
+    return settings.LANGUAGE_CODE
+
+
+def get_language_choices():
+    return settings.LANGUAGES
+
+
+def get_user_language_choices():
+    return settings.USER_LANGUAGES
```
n0ctua: It might be safer to use `getattr(settings, "USER_LANGUAGES", settings.LANGUAGES)` here so this defaults to LANGUAGES if the setting is not there in a downstream project for some reason.

froide/helper/context_processors.py
```diff
@@ -17,6 +17,7 @@ def site_settings(request):
             request, "LANGUAGE_CODE", settings.LANGUAGE_CODE
         ),
         "DEFAULT_LANGUAGE_CODE": settings.LANGUAGE_CODE,
+        "USER_LANGUAGES": settings.USER_LANGUAGES,
```
n0ctua: Same as in `froide/helper/language.py`

n0ctua (approved):
Looks good to me. I just added some questions and possible pitfalls in the comments you might want to check out.

froide/account/forms.py
stefanw: From the database point of view it's just a string. Choices are not validated on the DB level. I think having a no longer available language in the database is a problem that can be solved per deployment and doesn't need addressing here.

froide/foirequest/forms/request.py
noeddl: You're right, the field is recording the active locale, not offering a choice, so it should use `settings.LANGUAGES`.

PR #1248, opened 2026-04-16
prevent duplicate items in search registry

krmax44:
- **➕ add django-stubs-ext**
- **🐛 prevent searches from being registered multiple times**


stefanw:
This needs changes in all call sites, I'd go for an optional `name` keyword parameter in `search_registry.register`. So far only CMS searches are duplicated (due to the way they are registered), so only those need to be deduplicated.

stefanw (approved):

PR #1253, opened 2026-04-20
🐛 don't show public body proposals in search results

krmax44:
Currently, proposed public bodies are visible in the request form, which results in a 403 error screen. This PR prevents unconfirmed public bodies from being added to the search index.

They subsequently are also no longer visible in the public body search views, but I think that might be sensible too.

### DIFF
--- a/froide/publicbody/tests/test_search.py
+++ b/froide/publicbody/tests/test_search.py
@@ -4,6 +4,7 @@ import pytest
 
 from froide.foirequest.tests.factories import rebuild_index
 from froide.georegion.factories import GeoRegionFactory
+from froide.helper.search.signal_processor import realtime_search
 
 from ..factories import (
     CategoryFactory,
@@ -370,3 +371,33 @@ class TestPublicBodySearchAPI:
         view_names = {obj.name for obj in view_response.context["object_list"]}
 
         assert api_names == view_names
+
+
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.xdist_group(name="sequential")
+def test_proposals_not_in_search(world, client):
+    rebuild_index()
+
+    api_url = reverse("api:publicbody-search")
+
+    # make sure proposed public bodies are not indexed and returned in search api
+    with realtime_search():
+        pb_proposal = PublicBodyFactory.create(name="Indextest-1", confirmed=False)
+        pb_confirmed = PublicBodyFactory.create(name="Indextest-2", confirmed=True)
+
+    api_response = client.get(api_url, {"q": "Indextest"})
+    assert api_response.status_code == 200
+    data = api_response.json()
+    assert len(data["objects"]) == 1
+    assert data["objects"][0]["id"] == pb_confirmed.id
+
+    # confirming a proposal should make it appear in the index via the save signal
+    with realtime_search():
+        pb_proposal.confirmed = True
+        pb_proposal.save()
+
+    api_response = client.get(api_url, {"q": "Indextest"})
+    assert api_response.status_code == 200
+    data = api_response.json()
+    result_ids = {obj["id"] for obj in data["objects"]}
+    assert result_ids == {pb_proposal.id, pb_confirmed.id}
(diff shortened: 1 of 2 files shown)


froide/publicbody/tests/test_search.py
```diff
@@ -370,3 +370,22 @@ def test_api_and_view_return_same_results(self, client, publicbody_data, params)
...
+    rebuild_index()
+
+    api_url = reverse("api:publicbody-search")
+    api_response = client.get(api_url, {"q": "Indextest"})
+
+    assert api_response.status_code == 200
+
+    data = api_response.json()
+    assert len(data["objects"]) == 1
+    assert data["objects"][0]["id"] == pb_unconfirmed.id
```
noeddl, line 391: I noticed that the test passed even without `should_index_object`, because in tests the signal path is disabled and `rebuild_index` goes through a manager that already filters unconfirmed bodies, so `should_index_object` only gets to see confirmed bodies and its filtering behavior is never observed. I suggest to update the test to call `PublicBodyDocument.update()` directly, which mirrors the production signal path - see the test comment for details.

```suggestion
def test_proposals_not_in_search(client):
    # Make sure proposed public bodies are not indexed and returned in search api.
    #
    # rebuild_index goes through PublicBodyDocument.get_queryset -> the
    # PublicBodyManager, which already filters confirmed=True, so it cannot
    # exercise should_index_object. In tests, post_save signal indexing is also
    # disabled (BaseSignalProcessor). We therefore hand both bodies directly to
    # PublicBodyDocument.update(), which is the code path that consults
    # should_index_object, mirroring what RealTimeSignalProcessor does in prod.
    rebuild_index()

    pb_unconfirmed = PublicBodyFactory.create(name="Indextest-1", confirmed=False)
    pb_confirmed = PublicBodyFactory.create(name="Indextest-2", confirmed=True)

    PublicBodyDocument().update([pb_unconfirmed, pb_confirmed])
    publicbody_index.refresh()

    api_url = reverse("api:publicbody-search")
    api_response = client.get(api_url, {"q": "Indextest"})

    assert api_response.status_code == 200

    data = api_response.json()
    assert len(data["objects"]) == 1
    assert data["objects"][0]["id"] == pb_confirmed.id
```

noeddl (approved):

froide/publicbody/tests/test_search.py
noeddl: Ah, the updated tests also needs additional imports:
`from ..documents import PublicBodyDocument, index as publicbody_index`

noeddl: Now I discovered that this can be solved in an easier way by using [realtime_search](https://github.com/okfde/froide/blob/8da4c8e07ae92f674ce717cd0d0b3b3f610714da/froide/helper/search/signal_processor.py#L13-L18) (import with `from froide.helper.search.signal_processor import realtime_search`). Here is my new suggestion, including an additional check for confirming a previously unconfirmed request:

```python
@pytest.mark.django_db(transaction=True)
@pytest.mark.xdist_group(name="sequential")
def test_proposals_not_in_search(client):
    rebuild_index()

    api_url = reverse("api:publicbody-search")

    # make sure proposed public bodies are not indexed and returned in search api
    with realtime_search():
        pb_proposal = PublicBodyFactory.create(name="Indextest-1", confirmed=False)
        pb_confirmed = PublicBodyFactory.create(name="Indextest-2", confirmed=True)

    api_response = client.get(api_url, {"q": "Indextest"})
    assert api_response.status_code == 200
    data = api_response.json()
    assert len(data["objects"]) == 1
    assert data["objects"][0]["id"] == pb_confirmed.id

    # confirming a proposal should make it appear in the index via the save signal
    with realtime_search():
        pb_proposal.confirmed = True
        pb_proposal.save()

    api_response = client.get(api_url, {"q": "Indextest"})
    assert api_response.status_code == 200
    data = api_response.json()
    result_ids = {obj["id"] for obj in data["objects"]}
    assert result_ids == {pb_proposal.id, pb_confirmed.id}
```

PR #1259, opened 2026-04-30
🗃️ ensure publicbody slugs are unique

krmax44:
⚠️this  migration might require some manual intervention on some froide instances (i.e. clean-up of duplicate public bodies)


stefanw:
The PB URLs use the ID as a lookup (slug is just for decoration). If it weren't for make request URLs that historically allow PB slugs, this would not be necessary. 

Putting slugs in the make request URLs looked cute, but implicitly requiring uniqueness was a mistake. Unique slugs scale badly to authorities in multiple jurisdictions and become tricky if we ever go forward with translated fields of public bodies (e.g. for EU level). There are lots of authorities named the same way just located in a different jurisdiction or region. The problem will be to make slugs unique ideally without just tacking ugly numbers to the end.

So maybe a plan to remove the slug from the make request URL? Or are there other places where the unique slug is necessary?

PR #1279, opened 2026-06-04
prevent multiple make foirequest form submissions

krmax44:
I think handling this on the frontend is fine. Ideally, we have some kind of idempotency check (has the same user filed an FOI request with the same subject to the same public body in the last 5 minutes), but that seems difficult to implement without race conditions.

### DIFF
--- a/froide/foirequest/views/make_request.py
+++ b/froide/foirequest/views/make_request.py
@@ -284,6 +284,7 @@ class MakeRequestView(FormView):
                     "Are you sure you want to discard your current selection?"
                 ),
                 "loadMore": _("load more..."),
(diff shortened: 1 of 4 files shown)


stefanw (approved):

frontend/javascript/components/makerequest/review-request.vue
```diff
@@ -280,6 +280,10 @@ export default {
     hasProof: {
       type: Boolean,
       required: true
+    },
+    submitting: {
+      type: Boolean,
+      required: true
```
noeddl: `submitting` is defined as required but never used in this file?

PR #1306, opened 2026-08-17
MessageKind - allow different transports for initial message

fin:
For FragDenStaat.at, I need a way to send Requests initial messages via non-email channels. This is an attempt to make this possible by extending the existing MessageHandler framework. The idea (which is roughed out in the tests) is that the initial FOIMessage.kind can be determined using a MessageHandler classmethod receiving the request. This can check f.e. a different model to see if a PublicBody or a certain kind of request should be handled using non-email channels.

Happy to hear your feedback!


stefanw (approved):

froide/foirequest/models/message.py
```diff
@@ -31,7 +31,7 @@
 class FoiMessageManager(models.Manager):
     def get_throttle_filter(self, queryset, user, extra_filters=None):
         qs = queryset.filter(
-            sender_user=user, is_response=False, kind=MessageKind.EMAIL
+            ~models.Q(kind=MessageKind.UPLOAD), sender_user=user, is_response=False
```
krmax44: Maybe a bit clearer like that:

```suggestion
            sender_user=user, is_response=False
          ).exclude(kind=MessageKind.UPLOAD)
```

fin:
ok if i merge this? i have push permissions, but i don't know if you have any other procedures i should follow these days (:

PR #1313, opened 2026-09-02
Improve accessibility of account management

flooguu:


froide/account/templates/account/settings.html
```diff
@@ -66,7 +66,7 @@ <h5 class="card-header">{% trans "Change details" %}</h5>
         </div>
     </div>
     <div id="settings" class="card mb-3 mt-5">
-        <h5 class="card-header">{% trans "Account settings" %}</h5>
+        <h5 class="card-header">{% trans "Account Settings" %}</h5>
```
noeddl, line 69: I think "Account settings" with small "s" is actually "correct" here because it's consistent with the spelling of the other card headers in this view ("Change details", "Public profile", etc.).

noeddl (approved):

PR #1314, opened 2026-09-02
Fix dangling and missing aria-describedby references in form fields

noeddl:
Django sets `aria-describedby="<auto_id>_helptext <auto_id>_error"` on form widgets [automatically](https://docs.djangoproject.com/en/5.2/ref/forms/api/#django.forms.BoundField.aria_describedby), but our Bootstrap field templates rendered help texts and error messages without any id, so every described field pointed at elements that didn't exist. For widgets that use a fieldset (radio groups, multiple checkboxes) [Django leaves the attribute to the template](https://github.com/django/django/blob/4ea267661b260ea0d0c87e9dcb99d70037c6f2fc/django/forms/boundfield.py#L294), and ours never set it, so help text and errors weren't announced at all.                                                                                            
                                                                                                                                                                                                                                                       
### Changes                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                          
  - Factor out a `field_footer.html` partial shared by the horizontal, stacked and inline field templates, replacing four near-identical copies of the help-text/error markup.                                                                                               
  - Give help texts and errors the ids Django expects and set `aria-describedby` on the `<fieldset>` for widgets that use one. Radio/checkbox groups in the stacked and inline templates now use a real `<fieldset>/<legend>` instead of a label plus loose checkboxes.                                                                                                                                                                                                                                       
  - Show errors that Bootstrap kept hidden: `.invalid-feedback` stays `display: none` unless a preceding sibling has `.is-invalid`, which never happens for controls nested in a `.form-check`. The field footer partial adds `d-block` so error messages are actually displayed.
  - Fix the `hide_project_duplicates` checkbox in `_request_search.html`, whose help text is written out by hand rather than through the field templates and so was missing the id too.               
  
### Tests                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                           
  - New `froide/helper/tests/test_form_helper.py`: renders every widget type through all three template variants and asserts that help text and errors are present, referenced, displayed, and that no `aria-describedby` points at a missing id.                              
  - Extend end-to-end a11y test coverage for pages that previously had these problems.                                                                                                                                                                                                          
  - Update a11y snapshots to capture the fixed violation (`aria-valid-attr-value`).

Closes https://github.com/fragdenstaat/issues/issues/380

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ The end-to-end tests check the pages they visit with [axe-core](https://github.c
 
 ```shell
 # Record the current findings as the new baseline, then review the diff and commit
+# (Recording stops at the first snapshot of a test, so repeat until the tests pass)
 pytest froide/tests/live --force-regen
 # Show the full axe report for every finding, recorded or not
 pytest froide/tests/live --a11y-strict
--- a/froide/conftest.py
+++ b/froide/conftest.py
@@ -64,6 +64,11 @@ def dummy_user():
     yield UserFactory(username="dummy")
 
 
+@pytest.fixture
+def dummy_staff_user():
+    yield UserFactory(username="dummy_staff", is_staff=True)
+
+
 @pytest.fixture()
 def request_throttle_settings(settings):
     froide_config = settings.FROIDE_CONFIG
--- a/froide/foirequest/templates/foirequest/_request_search.html
+++ b/froide/foirequest/templates/foirequest/_request_search.html
@@ -76,7 +76,10 @@
                                 {{ form.hide_project_duplicates.label }}
                             </label>
                             {% if form.hide_project_duplicates.help_text %}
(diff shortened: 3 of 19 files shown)


flooguu (approved):
looks good to me 👍

don't 100% understand the changes to [froide/tests/live/test_a11y.py‎](https://github.com/okfde/froide/pull/1314/changes#diff-c889334783ca98abc09f22fd3ef03189eeea00f15c585a03ef1b1e06c75190f4) in relation to the PR.

noeddl:
Thanks for the review! My main intention with the changes in `test_a11y.py` was to extend the coverage of the a11y tests by adding some more pages that were affected by the problem fixed in this PR. But because some of these pages require to be logged in etc., I had to add some additional stuff around rather than just adding the pages to the existing tests. But I see that this might be a bit out of scope here. :sweat_smile:

PR #1316, opened 2026-09-04
Improve a11y on request page

flooguu:
- Tooltips on buttons (e.g. "Was bedeutet das?") are now reachable via keyboard and exposed to screen readers
- Collapsed messages now have a visually apparent toggle button, which also collapses the message again
- The "Details" control on messages is now a semantic <button> instead of a link, since it doesn't navigate to another page; it sets aria-expanded="true" when active so assistive tech is informed that the expanded content appeared
- The request-information timeline now has a heading, so it can be understood during non-visual use

### DIFF
--- a/froide/foirequest/templates/foirequest/body/message/_badge.html
+++ b/froide/foirequest/templates/foirequest/body/message/_badge.html
@@ -1,0 +1,7 @@
+<span class="alpha-message__badge {{ badge_class }}"
+      data-bs-toggle="tooltip"
+      title="{{ label }}"
+      role="img"
+      aria-label="{{ label }}">
+    <span class="fa {{ icon_class }}" aria-hidden="true"></span>
+</span>
--- a/froide/foirequest/templates/foirequest/body/message/message.html
+++ b/froide/foirequest/templates/foirequest/body/message/message.html
@@ -18,7 +18,8 @@
         <button type="button"
                 class="alpha-message__expand-toggle d-print-none"
                 aria-expanded="{% if message.timestamp == object.last_message %}true{% else %}false{% endif %}"
-                aria-controls="{{ message.get_html_id }}-body">
+                aria-controls="{{ message.get_html_id }}-body"
+                title="{% trans "Toggle message" %}">
             <i class="fa fa-angle-down alpha-message__expand-chevron"
                aria-hidden="true"></i>
             <span class="visually-hidden">{% trans "Toggle message" %}</span>
--- a/froide/foirequest/templates/foirequest/body/message/message.html
+++ b/froide/foirequest/templates/foirequest/body/message/message.html
@@ -34,18 +34,12 @@
                 {# icons & timestamp #}
                 <div class="d-flex flex-nowrap">
                     {% if message.all_attachments %}
-                        <span class="alpha-message__badge"
-                              data-bs-toggle="tooltip"
-                              title="{% trans 'Has attachments' %}">
-                            <span class="fa fa-paperclip"></span>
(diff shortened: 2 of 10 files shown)


froide/foirequest/templates/foirequest/body/message/_badge.html
```diff
@@ -0,0 +1,7 @@
+<span class="alpha-message__badge {{ badge_class }}"
```
noeddl, line 1: The badges are still not reachable via keyboard. Is this deliberate? (I think the question whether they should be came up during our pair programming session but I don't remember if we agreed on anything then.)

froide/foirequest/templates/foirequest/body/message/message.html
```diff
@@ -99,17 +108,21 @@
                     </span>
                 {% endif %}
                 <!-- delivery details link-->
-                <a href="#" class="alpha-message__meta-toggle text-body">
+                <button type="button"
+                        class="alpha-message__meta-toggle text-body"
+                        aria-expanded="false"
+                        aria-controls="{{ message.get_html_id }}-meta">
```
noeddl: If I see this correctly, `aria-controls` currently links to nowhere. I think the id has to be set on the outer div in meta_container.html.

froide/foirequest/templates/foirequest/body/timeline.html
```diff
@@ -1,6 +1,7 @@
 {% load i18n %}
 {% load content_helper %}
 {% block foirequest_timeline %}
+    <h4>{% trans "Timeline of this request" %}</h4>
```
noeddl: - The heading looks quite large, maybe it could be styled a bit smaller (e.g. with `class="h6"`). Maybe the text could be shortened as well to just "Timeline/Verlauf" (or maybe "Nachrichtenverlauf").
- The heading levels on this page need to be reworked anyway, but I think at the current stage `h3` would be semantically correct.
- On mobile, the heading is still visible without a timeline underneath.
- When scrolling, the heading disappears while the timeline underneath is sticky.

noeddl (commented):
I like that it's much clearer now that the messages can be expanded.

Little nitpick: The alignment of the chevron next to "Details" looks a bit off when the detail panel is collapsed (on Firefox).

Image

frontend/javascript/components/request/Message.ts
```diff
@@ -15,13 +16,17 @@ export default class Message {
...
+      '.alpha-message__expand-toggle'
+    )
 
     // event listeners
 
     // header row
     element
       .querySelector('.alpha-message__head')
       ?.addEventListener('click', this.onHeadClick.bind(this))
+    this.expandToggle?.addEventListener('click', this.onHeadClick.bind(this))
```
noeddl:
```suggestion
```
The button doesn't need its own click listener as it's inside `.alpha-message__head`, whose handler already toggles. Enter/Space on a button propagates there too. That way, the newly added `stopPropagation()` in `onHeadClick` below is not needed either. `onHeadClick` is also the message head's handler, so it applies to every click in the message header, not just the button. With `stopPropagation()` those clicks no longer reach `document`, so Bootstrap dropdowns no longer close. Example: open the download menu in the info box on the right side, then click a message header -> the download menu stays open (previously, it closed).

froide/foirequest/templates/foirequest/body/message/_badge.html
flooguu: Testing with VoiceOver on Mac the Badge is reached and has a proper label

PR #1322, opened 2026-09-11
♿️ Fix heading structure across templates

noeddl:
Fixes the heading structure across templates, changing the document outline with as little visual change as possible:
  - every page gets exactly one `<h1>`,
  - heading levels no longer skip steps,
  - text that is not a section label is no longer a heading.
                                                                
  Visual appearance is preserved with Bootstrap's heading classes (e.g. `<h1 class="h2">`). Where an element is needed in the outline but shouldn't look like a heading, it becomes a heading with the new `.heading-reset` class, which renders it like regular paragraph text.
  
Other changes:
  - The list item partials (`request_item_inner.html`, `publicbody_item.html`, `document/result_item.html`) accept a `heading_tag` parameter, since they are included at different outline depths. It defaults to `<h5>`, so existing includes in downstream themes are unchanged.
  - The footer styles apply to all heading levels and `.h1`–`.h6` classes instead of just `<h6>`, so any heading size can be used in the footer.
  
Most pages can be reviewed using the fragdenstaat_de dev setup. The exceptions are the index page and the footer, which are specific to froide and have to be checked in a standalone froide setup.

### DIFF
--- a/froide/account/templates/account/confirmed.html
+++ b/froide/account/templates/account/confirmed.html
@@ -15,7 +15,7 @@
         </div>
         <div class="row mt-5 justify-content-center">
             <div class="col-md-8 text-center">
-                <h2>{% trans "Your email address is now confirmed!" %}</h2>
+                <h1 class="h2">{% trans "Your email address is now confirmed!" %}</h1>
                 {% if foirequest %}
                     <p class="lead">
                         {% blocktrans trimmed with title=foirequest.title url=foirequest.get_absolute_url %}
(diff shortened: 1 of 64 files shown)


flooguu:
Bildschirmfoto 2026-09-18 um 11 05 43

This might be out of scope or I am diagnosing this wrong, but in the "Create request" flow (e.g. http://localhost:8000/anfrage-stellen/#step-4 ) there seems to be no heading 1 from step 2 onwards

flooguu (approved):

noeddl:
Good catch. Again material for a new issue ...

PR #1331, opened 2026-09-14
🔧 dependabot: ignore major updates

krmax44:


pymnh (approved):

.github/dependabot.yml
```diff
@@ -6,6 +6,10 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    allow:
+      update-types:
+        - version-update:semver-patch
+        - version-update:semver-minor
```
pymnh, line 12:
```suggestion
      - dependency-type: all
        update-types:
          - version-update:semver-patch
          - version-update:semver-minor
```

.github/dependabot.yml
```diff
@@ -19,6 +23,10 @@ updates:
...
```
pymnh, line 29:
```suggestion
      - dependency-type: all
        update-types:
          - version-update:semver-patch
          - version-update:semver-minor
```

pymnh (changes requested):
`allow` requires a list, not an object, see
https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#allowing-specific-semantic-versioning-levels-for-updates

